Issue #321, opened 2026-04-14
[HIGH] Engine setup-failure path skips checkpoint, losing state on crash before final persist


vanyastaff:
## Summary
`Engine::run_frontier` does not checkpoint execution state when node setup fails before spawn (e.g. parameter resolution error). Runtime failures are checkpointed immediately, but setup failures are only persisted at final execution checkpoint.

## Impact
- Crash/restart window can lose failed-node state and edge-resolution decisions.
- Resume may re-run nodes that had already failed during setup, causing duplicate side effects or divergent control flow.
- Durability semantics differ by failure phase (setup vs runtime), which is unexpected and unsafe.

## Failure Scenario
1. A node fails in `spawn_node` during parameter resolution.
2. Engine marks node failed and applies error strategy in-memory (`handle_node_failure`), but does not call `checkpoint_node`.
3. Process crashes before final state checkpoint.
4. On resume, persisted execution state does not reflect that setup failure handling occurred.

## Evidence
- `crates/engine/src/engine.rs`:
  - Setup-failure path: `spawned == false` branch calls `handle_node_failure(...)` but has no checkpoint call.
  - Runtime-failure path (`join_set` result `Err`) explicitly checkpoints right after `handle_node_failure(...)`.
- This creates an asymmetry in persistence guarantees for logically similar node failures.

## Why Tests/Lints Missed It
- Existing persistence tests assert final state, not mid-execution crash consistency.
- No fault-injection test that crashes between setup failure handling and final checkpoint.

## Proposed Fix
- After setup failure + `handle_node_failure(...)`, call the same `checkpoint_node(...)` used by runtime failures.
- Add regression test that simulates crash/restart after setup failure and verifies resumed state does not re-execute failed node incorrectly.

## Acceptance Criteria
- Setup-failure path persists node state/outputs/edge effects before loop continues or aborts.
- Crash between setup failure and final completion does not lose that failure handling on resume.

## Related Files
- `crates/engine/src/engine.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

Issue #311, opened 2026-04-14
[HIGH] resume_execution drops original workflow input (replays with Null)


vanyastaff:
## Summary
`resume_execution` in `nebula-engine` does not restore original workflow trigger input. It hardcodes `workflow_input = Null` when re-entering the frontier loop.

## Impact
- Crash-recovered entry nodes can re-run with different input than the original execution.
- Non-idempotent entry actions may produce incorrect side effects (wrong API payloads, wrong routing keys, missing auth context).
- Behavior diverges between fresh execution and resumed execution, making recovery non-deterministic.

## Failure Scenario
1. Workflow starts with non-null trigger input (e.g. webhook payload).
2. Process crashes before an entry node reaches terminal state (so idempotency does not short-circuit it).
3. `resume_execution` rebuilds frontier and dispatches that entry node.
4. Instead of original trigger payload, node receives `Null`.
5. Node output / downstream graph diverges from original intended run.

## Evidence
- `crates/engine/src/engine.rs` in `resume_execution`:
  - TODO explicitly states original workflow input is not persisted.
  - Code sets `let workflow_input = serde_json::Value::Null;`
  - That value is passed to `run_frontier(...)`.

## Why Tests/Lints Missed It
- Existing tests mostly validate happy-path execution and generic recovery mechanics.
- No regression test currently simulates crash recovery where an entry node is *not* terminal and depends on original trigger payload.
- Static analysis/lints cannot detect semantic state-loss across restart boundaries.

## Proposed Fix
- Persist original trigger/workflow input in `ExecutionState` (or linked durable record) at execution start.
- Restore persisted input in `resume_execution` instead of forcing `Null`.
- Add recovery regression test that:
  - starts execution with non-null input,
  - simulates crash before entry completion,
  - resumes execution and asserts resumed node sees original input.

## Acceptance Criteria
- Resumed execution uses byte-equivalent original input for any re-dispatched entry node.
- Recovery behavior is deterministic with and without process restart.
- Regression test fails on current `Null` behavior and passes after fix.

## Related Files
- `crates/engine/src/engine.rs`
- `crates/execution/src/state.rs`
- `crates/storage/src/backend/pg_execution.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

Issue #301, opened 2026-04-14
[MEDIUM] engine run_frontier: panicked NodeTask reports NodeId::new() — cannot identify which node panicked


vanyastaff:
## Summary

When a spawned `NodeTask` panics, `run_frontier` discards the panicking node's identity and returns a **freshly-generated** `NodeId` as the error's node reference:

```rust
// crates/engine/src/engine.rs ~1134
Err(join_err) => {
    tracing::error!(?join_err, \"node task panicked\");
    cancel_token.cancel();
    return Some((NodeId::new(), join_err.to_string()));
}
```

The returned `NodeId` has no relationship to the panicked task. Upstream callers log this as \"node X failed\" where X is a random UUID not present in the workflow definition.

## Effect

- **Diagnostics destroyed.** The only signal in logs is \"node `<random uuid>` panicked\", with no way to correlate to the real node. The engine's own event stream emits `ExecutionEvent::NodeFailed` with the same bogus ID.
- **Metrics broken.** Alert rules that aggregate by node ID either see a cardinality explosion (a new \"failed node\" on every panic) or silently drop the counter.
- **Cleanup cannot happen.** The engine can't mark the actually-panicked node as Failed because it doesn't know which one it is. The node stays in whatever state it was in (usually Running) even though its task is gone, so reconciliation / graceful shutdown sees a ghost Running node and waits forever.

## Why the bug exists

`tokio::task::JoinSet::join_next()` returns `Result<T, JoinError>`. On panic, the `Err(JoinError)` does not carry the task's associated data (`T`) — `T` here is `(NodeId, Result<...>)`, so the `NodeId` component is inaccessible once the task aborts.

The engine needs a side-table to recover the identity.

## Fix sketch

`JoinSet::spawn` returns a `tokio::task::AbortHandle` (or in `JoinSet` specifically, `spawn` gives you a `Id` via `spawn_by_id` / `tokio::task::Id`). Keep a `HashMap<tokio::task::Id, NodeId>` alongside `join_set`:

```rust
let mut task_ids: HashMap<tokio::task::Id, NodeId> = HashMap::new();

let abort = join_set.spawn(NodeTask { ... }.run());
task_ids.insert(abort.id(), node_id);

// Later:
match join_set.join_next_with_id().await {
    Some(Ok((task_id, (node_id, result)))) => {
        task_ids.remove(&task_id);
        // happy path: use the returned node_id (which matches our map)
    }
    Some(Err(join_err)) => {
[...]

Issue #300, opened 2026-04-14
[MEDIUM] engine: spawn_node's chained transition_to calls swallow invalid transitions — node runs in inconsistent state


vanyastaff:
## Summary

`WorkflowEngine::spawn_node` advances a node through Ready → Running via three discarded results:

```rust
// crates/engine/src/engine.rs ~1196
if let Some(ns) = exec_state.node_states.get_mut(&node_id) {
    let _ = ns.transition_to(NodeState::Ready);
    let _ = ns.transition_to(NodeState::Running);
}
```

and the param-resolution failure path does the same:

```rust
// ~line 1185
if let Some(ns) = exec_state.node_states.get_mut(&node_id) {
    let _ = ns.transition_to(NodeState::Ready);
    let _ = ns.transition_to(NodeState::Running);
    let _ = ns.transition_to(NodeState::Failed);
    ns.error_message = Some(e.to_string());
}
```

Each `transition_to` validates the source state per `crates/execution/src/transition.rs`. Every `let _` discards both the error *and* the invariant that \"the state is in the expected place\". If the node is already `Failed` (retry path) or `Cancelled`, the first `transition_to(Ready)` fails silently, the follow-up transitions then fire from whatever state the first one left things in, and the code still proceeds to **spawn the task**.

## Concrete failure

Consider a retry of a failed node:

1. Node state is `Failed` (from the previous attempt).
2. `run_frontier` puts the node back on `ready_queue` (retry policy).
3. `spawn_node` runs the chained transitions:
   - `Failed → Ready` — invalid per `transition.rs:44-58` (Failed only goes to Retrying / Cancelled). `let _` swallows the error.
   - `Ready → Running` — source is still `Failed`, same mismatch.
4. The node's `ns.state` remains `Failed`, but `spawn_node` returns `true` and the task executes.
5. On success, `mark_node_completed` attempts `Failed → Completed`, another invalid transition.

The node runs in a state the state machine says is impossible. Anything reading `node_states[node_id].state` during the window reports `Failed` for a node that is actively running. Dashboards, reconciliation jobs, and the replay partitioner all act on stale data.

The param-failure path is worse — it forces Ready → Running → Failed, which in isolation looks defensible, but because the three `let _` discard errors, a node that was already terminal (Cancelled) ends up with `error_message` clobbered and `state` unchanged.

## Related

- `#255` — `ExecutionState::set_node_state` bypasses version bump / transition validation. Same pattern, different method.
[...]

PR #346, opened 2026-04-14
fix(engine): batch 2 execution-state correctness (#299, #300, #301, #311, #321)

### LINKED ISSUE #299: [HIGH] check_and_apply_idempotency reconstructs ActionResult as Success — Branch/Route/MultiOutput semantics lost on resume
## Summary

When the engine hits an already-idempotent node on resume, `check_and_apply_idempotency` loads the persisted output and synthesizes a fake result to drive edge evaluation:

```rust
// crates/engine/src/engine.rs ~1343-1357
outputs.insert(node_id, output_value.clone());

let fake_result = ActionResult::success(output_value);
process_outgoing_edges(
    node_id,
    Some(&fake_result),
    None,
    graph,
    activated_edges,
    resolved_edges,
    required_count,
    ready_queue,
    exec_state,
);
```

The fake result is always `ActionResult::Success { output }`. But the **real** result from the original execution may have been:

- `Branch { selected, output, alternatives }` — downstream edges with `branch_key` must only fire when `key == selected`.
- `Route { port, data }` — downstream edges must only fire when `from_port == port`.
- `MultiOutput { outputs, main_output }` — downstream edges select by port name into `outputs`.
- `Skip { .. }` — no edge fires at all.
- `Wait { .. }` — node is not complete yet, the engine should re-enqueue later.

## Effect

Replaying or resuming an execution whose branches were selected by a node returning `Branch { selected: \"approved\" }`:

1. On first run: only the `approved`-tagged edge activated, `rejected` was pinned as a dead alternative.
2. On resume: `check_and_apply_idempotency` fakes the result as `Success`, passes it to `process_outgoing_edges`, and the branch-matching logic in `evaluate_edge`:

```rust
// ~line 1692
if let Some(ActionResult::Branch { selected, .. }) = result
    && let Some(ref key) = conn.branch_key
    && key != selected
{
    return false;
}
```

never matches because `result` is `Success` instead of `Branch`. **Every** branch edge activates, regardless of `branch_key`. The resumed workflow executes both `approved` AND `rejected` paths — potentially duplicating side effects on the \"rejected\" branch that the original never ran.

Same failure mode for `Route` (port mismatch filter never fires → all port edges activate) and `MultiOutput` (similarly).

## Location

`crates/engine/src/engine.rs` — `check_and_apply_idempotency` around line 1346, interacting with `evaluate_edge` around line 1654.

## Why this is HIGH

- Silent divergence between first-run and resume semantics, exactly the class of bug that survives unit tests.
- Reachable via any workflow that uses `Branch` / `Route` / `MultiOutput`, which are the engine's entire conditional-routing toolkit.
- Resume is the normal recovery path after any restart, so the bug activates under the conditions the feature is designed for.

## Fix sketch

The engine must persist the **structural variant** of `ActionResult`, not only the primary output. Options:

1. **Persist the full `ActionResult<Value>`.** Check `ActionResult` already derives `Serialize/Deserialize` — if not, add it to the variants that carry routing information. Store the whole thing in the repo next to the raw output, load it on the idempotency hit, pass it to `process_outgoing_edges` directly.

2. **Persist the `selected` branch / `port` / `outputs` keys.** Store a `PersistedResultShape` enum that the engine recomputes into a `fake_result` matching the original variant. Smaller on-disk footprint but more fragile.

3. **Persist the edge-activation decision itself.** Instead of replaying the ActionResult, persist which outgoing edges activated on the original run and feed that directly into `activated_edges` on resume. Skips the `evaluate_edge` step entirely for replayed nodes. Cleanest but requires a schema change to the edge-tracking store.

Option (1) is the smallest code delta and keeps `evaluate_edge` as the single source of truth. Option (3) is the most correct long-term.

## Severity

**HIGH** — correctness divergence between first run and resume, on a first-class feature (`Branch`/`Route`/`MultiOutput`). Especially dangerous because it reads as \"clean\" on normal tests (no resume) and fails silently when recovery kicks in.

vanyastaff:
## Summary

Five high-priority correctness bugs in `WorkflowEngine::run_frontier` and its persistence surface. Each was identified in the deep-review pass and tracked as its own issue.

- **#321** — `run_frontier`'s setup-failure branch now calls `checkpoint_node` (same ordering as the runtime-failure branch) and emits `NodeFailed`. Previously the in-memory state moved to `Failed` but the persisted state still described the node as `Pending`, so CAS readers and `resume_execution` saw a ghost `Pending` node.
- **#300** — `spawn_node` now drives state transitions through a typed `NodeExecutionState::start_attempt` helper that models the legal paths (`Pending → Ready → Running`, `Failed → Retrying → Running`, `Retrying → Running`) and returns `InvalidTransition` for anything else. `ExecutionState::start_node_attempt` wraps it with a parent version bump. On error the node is routed through the existing setup-failure path via `mark_setup_failed` instead of silently spawning a task on stale `Failed` state — the previous `let _ = ns.transition_to(Ready)` pattern swallowed the error.
- **#301** — Panicked `NodeTask`s no longer report `NodeId::new()`. `run_frontier` tracks a `HashMap<tokio::task::Id, NodeId>` alongside the `JoinSet`, switches to `join_next_with_id`, and looks up the real `NodeId` via `JoinError::id()`. Panicked nodes are marked `Failed`, checkpointed, and reported via `NodeFailed` with the real identity.
- **#311** — `ExecutionState` now carries an `Option<Value> workflow_input` (`#[serde(default)]` for legacy states). `execute_workflow` and `replay_execution` set it at execution start; `resume_execution` reads it back so entry nodes that crashed mid-run receive the original trigger payload instead of `Null`. Legacy states fall through to `Null` with a warn log.
- **#299** — `check_and_apply_idempotency` no longer synthesizes `ActionResult::Success` on replay. It loads the full persisted `ActionResult` via new `ExecutionRepo::save_node_result` / `load_node_result` hooks (default no-op so the Postgres backend keeps its current schema; `InMemoryExecutionRepo` overrides both) and hands it to `process_outgoing_edges`. On a persisted `Branch`/`Route`/`MultiOutput` the routing semantics survive idempotency replay; previously every branch edge fired unconditionally.

## Test plan

- [x] `cargo check -p nebula-engine -p nebula-execution -p nebula-storage`
- [x] `cargo clippy -p nebula-engine -p nebula-execution -p nebula-storage --all-targets -- -D warnings`
- [x] `cargo nextest run -p nebula-engine` — **91 passed** (1 skipped)
- [x] `cargo nextest run -p nebula-execution` — all passing (6 new state-machine tests)
- [x] `cargo nextest run -p nebula-storage` — all passing
- [x] `cargo test --doc -p nebula-engine -p nebula-execution -p nebula-storage`
- [x] `cargo +nightly fmt`

## New regression tests

In `crates/engine/src/engine.rs`:

- `setup_failure_checkpoints_execution_state` — #321
- `resume_restores_original_workflow_input` — #311
- `start_node_attempt_rejects_terminal_state` — #300
- `panicked_task_reports_real_node_id` — #301
- `idempotency_replay_preserves_branch_routing` — #299

In `crates/execution/src/state.rs`:

- `start_attempt_pending_path`
- `start_attempt_retry_path`
- `start_attempt_rejects_completed`
- `execution_state_start_node_attempt_bumps_version`
- `mark_setup_failed_records_error_and_bumps_version`
- `workflow_input_roundtrip_via_serde`
- `workflow_input_missing_field_deserializes_as_none`

Closes #299 #300 #301 #311 #321

---

> [!NOTE]
> **Medium Risk**
> Touches core scheduling/persistence paths (`run_frontier`, `ExecutionState`, `ExecutionRepo`) and changes failure handling and replay semantics; while well-tested, regressions could impact resume/idempotency correctness in production backends that don’t yet persist full results.
> 
> **Overview**
> Fixes several correctness gaps in `WorkflowEngine::run_frontier` around **state transitions, checkpointing, and replay**.
> 
> Setup-time failures (e.g., param resolution or missing node defs) now go through `ExecutionState::mark_setup_failed`, emit `NodeFailed`, and **checkpoint after failure handling** so persisted state matches in-memory outcomes. Node dispatch is tightened by introducing typed `start_attempt`/`start_node_attempt` transitions and treating invalid transitions as setup failures instead of spawning tasks on stale state.
> 
> Improves resilience and replay fidelity: panicked JoinSet tasks are now attributed to the **real `NodeId`** via `join_next_with_id` + a task-id map (and are marked/checkpointed as failed), `ExecutionState` now persists `workflow_input` for accurate `resume_execution`, and idempotency replay can load/persist the full `ActionResult` via new `ExecutionRepo::save_node_result`/`load_node_result` hooks (default no-op; implemented for `InMemoryExecutionRepo`) so Branch/Route/MultiOutput routing survives replay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 114765ec6cb8245bc3cd27cf2c41aa81eb0d9591. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.project/context/crates/engine.md
+++ b/.project/context/crates/engine.md
@@ -27,3 +27,5 @@ Workflow execution orchestrator — frontier-based DAG scheduler.
 <!-- reviewed: 2026-04-14 — #247 added Drop/Terminate gate to evaluate_edge with TODO(engine) for Phase 3 scheduler wiring -->
 
 <!-- reviewed: 2026-04-14 -->
+
+<!-- reviewed: 2026-04-13 — batch 2 (#299 #300 #301 #311 #321): spawn_node now uses typed start_node_attempt helper and routes invalid transitions through mark_setup_failed + setup-failure checkpoint path; panicked JoinSet tasks recover the real NodeId via a task-id side map (join_next_with_id); resume restores workflow_input from persisted ExecutionState (fallback Null+warn on legacy states); idempotency replay loads the full ActionResult via save_node_result/load_node_result so Branch/Route/MultiOutput routing survives replay. -->
--- a/.project/context/crates/execution.md
+++ b/.project/context/crates/execution.md
@@ -25,3 +25,5 @@ Execution state machine types — persistent state, journals, idempotency, plans
 <!-- reviewed: 2026-04-14 — #247 added ExecutionTerminationReason + ExecutionTerminationCode newtype and ExecutionResult::termination_reason field; Phase 0 wiring is partial (local edge gate only, full scheduler propagation deferred) -->
 
 <!-- reviewed: 2026-04-14 -->
+
+<!-- reviewed: 2026-04-13 — batch 2 (#300 #311): NodeExecutionState::start_attempt models Pending->Ready->Running / Failed->Retrying->Running / Retrying->Running as typed transitions; ExecutionState::start_node_attempt wraps it with parent-version bump. ExecutionState::mark_setup_failed handles param-resolution-time failures (Pending->Failed via override_node_state + error_message). ExecutionState::workflow_input is Option<Value> with #[serde(default)] so legacy stored states round-trip; set by engine at execution start, read by resume_execution. -->
--- a/.project/context/crates/storage.md
+++ b/.project/context/crates/storage.md
@@ -23,3 +23,5 @@ Storage trait abstraction — MemoryStorage for tests, PostgresStorage for produ
 
 ## Relations
 - Depends on nebula-core (IDs). Used by nebula-engine, nebula-api.
+
+<!-- reviewed: 2026-04-13 — batch 2 (#299): ExecutionRepo::save_node_result / load_node_result persist the full ActionResult<Value> next to the primary node output so idempotency replay can reconstruct Branch/Route/MultiOutput routing. Default trait impl is no-op (Ok(None)/Ok(())) so backends opt in; InMemoryExecutionRepo overrides both. Postgres backend keeps the default (no migration) — idempotency replay on Postgres falls back to synthesized Success with a warn log until the column is added. -->
--- a/crates/config/src/watchers/polling.rs
+++ b/crates/config/src/watchers/polling.rs
@@ -356,7 +356,10 @@ impl ConfigWatcher for PollingWatcher {
             // on timeout, abort explicitly so the runtime reclaims the task
             // rather than leaving a zombie behind.
             let abort_handle = handle.abort_handle();
-            if tokio::time::timeout(self.interval * 2, handle).await.is_err() {
+            if tokio::time::timeout(self.interval * 2, handle)
+                .await
+                .is_err()
+            {
                 nebula_log::warn!("Polling task did not exit within timeout; aborting");
                 abort_handle.abort();
             }
--- a/crates/credential/src/credentials/oauth2.rs
+++ b/crates/credential/src/credentials/oauth2.rs
@@ -381,12 +381,8 @@ impl Credential for OAuth2Credential {
                 let challenge = crate::crypto::generate_code_challenge(&verifier);
                 let state_token = crate::crypto::generate_random_state();
 
-                let url = oauth2_flow::build_auth_url(
-                    &config,
-                    client_id,
-                    &challenge,
-                    &state_token,
-                )?;
+                let url =
+                    oauth2_flow::build_auth_url(&config, client_id, &challenge, &state_token)?;
 
                 // `build_config` rejects missing `redirect_uri` for this
                 // grant type, so this `clone()` unwraps a value that was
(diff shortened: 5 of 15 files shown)


vanyastaff:
Conflicts with current `main` are structural, not just textual. PR #346 uses `JoinSet::join_next_with_id` + a parallel `task_nodes` map to track in-flight node IDs, while `main` (after #326/#394) restructured the same loop to race `join_set.join_next()` against a wall-clock deadline via `tokio::select!` with `abort_all()` on cancel/timeout. The two designs are not naively compose-able — combining them requires deciding whether to:

1. Keep main's wall-clock budget enforcement and add per-task ID tracking on top (likely `tokio::select!` arm uses `join_next_with_id`), or
2. Drop main's wall-clock arm and reintroduce PR #346's simpler structure (loses the budget-during-empty-queue guard).

This needs a human call, not a mechanical merge. Closing as needs-rework — please rebase against current `main` and pick option (1) or (2). Other improvements from PR #346 (the execution-state correctness around #299/#300/#301/#311/#321) can land as smaller follow-ups.

Issue #323, opened 2026-04-14
[HIGH] DaemonRuntime stop is delayed by restart_backoff (cancel not observed during sleep)


vanyastaff:
## Summary
`DaemonRuntime` is not cancellation-responsive during restart backoff. The loop sleeps with `tokio::time::sleep(config.restart_backoff).await` and does not select on cancel token.

## Impact
- `stop()` can block until full backoff delay elapses.
- Graceful shutdown latency becomes unbounded by configured backoff (potentially minutes).
- Can violate orchestrator shutdown deadlines and leave tasks running longer than expected.

## Failure Scenario
1. Daemon run fails and enters restart path.
2. Runtime starts `sleep(restart_backoff)`.
3. Operator calls `stop()` (or parent shutdown triggers cancellation) during sleep.
4. Daemon task does not exit until sleep completes.

## Evidence
- `crates/resource/src/runtime/daemon.rs`:
  - `daemon_loop()` checks `cancel.is_cancelled()` before/after run.
  - Backoff phase uses plain `tokio::time::sleep(config.restart_backoff).await`.
  - No `tokio::select!` on `cancel.cancelled()` during backoff wait.

## Why Tests/Lints Missed It
- Existing tests focus on restart and lifecycle transitions, not stop responsiveness under backoff.
- Static analysis cannot infer cooperative cancellation gaps inside await points.

## Proposed Fix
- Replace plain sleep with cancellation-aware wait:
  - `tokio::select! { _ = cancel.cancelled() => break, _ = sleep(backoff) => {} }`
- Add regression test: stop during backoff exits promptly.

## Acceptance Criteria
- Calling `stop()` during backoff returns promptly (well below configured backoff duration).
- Test covers cancellation while waiting between restart attempts.

## Related Files
- `crates/resource/src/runtime/daemon.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

vanyastaff:
Resolved by PR #399 (merged as f9e97738). The `biased select` in `daemon_loop` races `restart_backoff` against the per-run cancel token, so `stop()` returns promptly even mid-backoff. Covered by `stop_during_restart_backoff_returns_promptly` in `crates/resource/src/runtime/daemon.rs` (10s backoff + stop must return under 1s).

Issue #322, opened 2026-04-14
[HIGH] RecoveryGate probe window is not serialized, allowing thundering herd after backoff


vanyastaff:
## Summary
`Manager` checks `RecoveryGate` state with a read-only snapshot (`gate.state()`), and when backoff expires it allows **all** callers through instead of granting a single probe ticket. This defeats herd protection at the probe boundary.

## Impact
- After `retry_at`, concurrent callers can all hit a still-dead backend at once.
- Causes retry storms / connection spikes exactly during recovery windows.
- Increases tail latency and can delay backend recovery under load.

## Failure Scenario
1. Backend starts failing; gate transitions to `Failed { retry_at, ... }`.
2. Traffic continues and waits for retry window.
3. `retry_at` passes while many requests arrive concurrently.
4. `check_recovery_gate()` returns `Ok(())` for each caller (no CAS claim).
5. All callers execute real acquire against failing backend simultaneously.

## Evidence
- `crates/resource/src/manager.rs`:
  - `check_recovery_gate()` uses `gate.state()` and for expired `Failed` returns `Ok(())` directly.
  - It does not call `gate.try_begin()` to atomically claim probe ownership.
- `trigger_recovery_on_failure()` does use `try_begin()`, but only **after** the backend call already happened, so the spike is not prevented.
- `RecoveryGate` was designed specifically to serialize expensive recovery attempts.

## Why Tests/Lints Missed It
- No concurrent test around the `retry_at` boundary with many contenders.
- Lints/type checks cannot detect lock-free protocol violations or missing CAS claim.

## Proposed Fix
- Replace snapshot-only gate check with CAS-based claim flow:
  - callers should use `try_begin()` on `Failed`/`Idle` path;
  - only ticket holder probes backend;
  - non-holders get `AlreadyInProgress`/`RetryLater`.
- Wire ticket resolution to acquire result (`resolve` on success, `fail_transient/permanent` on failure class).

## Acceptance Criteria
- At probe boundary, at most one caller reaches backend while others are gated.
- Concurrency test verifies no probe herd when `retry_at` expires under parallel load.

## Related Files
- `crates/resource/src/manager.rs`
- `crates/resource/src/recovery/gate.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

vanyastaff:
[...]

Issue #318, opened 2026-04-14
[HIGH] DaemonRuntime lifecycle is not restart-safe (stale handle + cancelled token reuse)


vanyastaff:
## Summary
`DaemonRuntime` has broken lifecycle semantics:
1. `stop()` cancels a persistent token that is reused for future `start()` calls, making subsequent runs immediately exit.
2. Finished daemon tasks never clear `handle`, so later `start()` calls can fail with `"daemon is already running"` even when no task is alive.

## Impact
- Daemon resources become effectively one-shot in long-lived processes.
- Restart/recovery flows can silently fail after first stop or first natural daemon exit.
- Operators may see false "already running" errors and unavailable background services.

## Failure Scenario
1. Start daemon runtime successfully.
2. Either:
   - call `stop()`, then call `start()` again, or
   - daemon exits naturally (policy/restart budget), then call `start()` again.
3. New `start()` does not produce a functioning daemon:
   - reused cancelled token causes immediate loop exit, or
   - stale `handle` blocks start path with permanent error.

## Evidence
- `crates/resource/src/runtime/daemon.rs`:
  - `stop()` calls `self.cancel.cancel()` on runtime-level token.
  - `start()` derives child token from `self.cancel` without resetting/replacing it.
  - `start()` only checks `guard.is_some()` and never checks `JoinHandle::is_finished()`.
  - completed task does not clear `self.handle`.

## Why Tests/Lints Missed It
- No lifecycle tests for `start -> stop -> start` or `start -> natural exit -> start`.
- Type/lint checks do not reason about runtime token reuse or stale finished handles.

## Proposed Fix
- Make cancellation token per-run (or recreate runtime-level token before restart).
- In `start()`, treat finished handles as stopped (cleanup stale handle before guarding).
- Add lifecycle regression tests for both restart paths.

## Acceptance Criteria
- `start -> stop -> start` reliably runs daemon again.
- `start -> daemon exits -> start` works without false "already running".
- Tests cover token reset and stale-handle cleanup.

## Related Files
- `crates/resource/src/runtime/daemon.rs`
- `crates/resource/src/topology/daemon.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

vanyastaff:
[...]

Issue #314, opened 2026-04-14
[HIGH] RefreshCoordinator allows zero concurrency, causing refresh deadlock


vanyastaff:
## Summary
`RefreshCoordinator::with_max_concurrent(max)` accepts `max = 0` and constructs a semaphore with zero permits. Any winner path calling `acquire_permit()` then waits forever, effectively deadlocking credential refresh.

## Impact
- Refreshable credentials can stall indefinitely once refresh is needed.
- Waiters pile up behind a winner that never acquires a permit, causing auth degradation/outage.
- Under load this can cascade into widespread request failures while appearing as generic timeout behavior.

## Failure Scenario
1. Coordinator is constructed with `with_max_concurrent(0)` (misconfig or bad default propagation).
2. A credential enters refresh path and becomes `RefreshAttempt::Winner`.
3. Winner calls `acquire_permit().await` on a zero-permit semaphore.
4. Await never resolves; refresh never runs; in-flight entry remains occupied until task cancellation.
5. Subsequent callers become waiters and eventually time out/retry without progress.

## Evidence
- `crates/credential/src/refresh.rs`:
  - `with_max_concurrent(max)` directly calls `Semaphore::new(max)` without clamping/validation.
  - `acquire_permit()` awaits `acquire_owned()` unconditionally.
- No API guard or validation rejects zero.

## Why Tests/Lints Missed It
- Existing tests cover normal permit counts and breaker behavior, not `max = 0`.
- Lints and type checks cannot infer semantic deadlock from valid numeric input.

## Proposed Fix
- Enforce `max >= 1` at construction:
  - either clamp (`max.max(1)`) or return `Result` with validation error for zero.
- Add regression test asserting zero cannot create a deadlocking coordinator.
- If clamping is chosen, document effective value in API docs.

## Acceptance Criteria
- `with_max_concurrent(0)` no longer allows a coordinator that can deadlock refresh.
- Regression test covers zero-input behavior.
- Docs explicitly define valid concurrency range.

## Related Files
- `crates/credential/src/refresh.rs`
- `crates/credential/src/resolver.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

PR #371, opened 2026-04-14
fix(resource): batch 3a lifecycle hardening (#302, #318, #322, #323)

### LINKED ISSUE #302: [HIGH] Manager::graceful_shutdown proceeds to registry.clear() even when wait_for_drain timed out — surviving ResourceHandles see destroyed resources
## Summary

`Manager::graceful_shutdown` runs four phases sequentially: Signal → Drain → Clear → Await Workers. Phase 2 (`wait_for_drain`) only **logs a warning** on timeout; it does not abort the shutdown. Phase 3 then unconditionally calls `self.registry.clear()`, destroying the managed resources even though handles are still outstanding.

```rust
// crates/resource/src/manager.rs ~1203
pub async fn graceful_shutdown(&self, config: ShutdownConfig) {
    self.cancel.cancel();                           // Phase 1
    self.wait_for_drain(config.drain_timeout).await; // Phase 2: logs warn on timeout
    self.registry.clear();                           // Phase 3: unconditional
    if let Some(handle) = self.release_queue_handle.lock().await.take() {
        let shutdown_fut = ReleaseQueue::shutdown(handle);
        if tokio::time::timeout(Duration::from_secs(10), shutdown_fut).await.is_err() {
            tracing::warn!(\"... did not finish within 10s, continuing shutdown\");
        }
    }
}
```

```rust
// ~1291
if drained.is_err() {
    tracing::warn!(
        active_handles = tracker.0.load(AtomicOrdering::Relaxed),
        \"resource manager: drain timeout expired with handles still active\"
    );
}
```

Note: `wait_for_drain` has no return type. A timeout is indistinguishable from success to `graceful_shutdown`.

## Failure sequence

1. Operator issues graceful shutdown with `drain_timeout = 5s`.
2. One plugin's `ResourceHandle` is still held by a long-running HTTP call.
3. Phase 2 waits 5 s, logs `\"drain timeout expired with handles still active\"`, returns.
4. Phase 3 `registry.clear()` drops the `ManagedResource` inside the registry. The resource's internal state (pool, connections, cached state) is released.
5. The long-running HTTP call eventually completes. The `ResourceHandle` tries to release back to the pool via `ReleaseQueue`. The ReleaseQueue worker processes the task.
6. Depending on implementation, one of:
   - The release worker panics because its `ManagedResource` ref is gone.
   - The release is silently dropped and the resource (e.g. a DB connection) leaks outside the manager's tracking — the Drop impl fires outside the supervised teardown window.
   - Worse: the in-flight HTTP call itself touches state that the registry clear has just released, producing a use-after-logical-drop.

Regardless of which exact path hits, the operator sees \"graceful shutdown complete\" in logs and the process exits cleanly while real work is still happening.

## Why this matters

- Graceful shutdown is the feature that *prevents* the symptoms above. The current implementation gives the opposite guarantee: timeout → silent force-clear, with no caller-visible error.
- `ShutdownConfig::drain_timeout` has default 30 s. Under a tight orchestrator shutdown window (k8s `terminationGracePeriodSeconds = 30`) the orchestrator will `SIGKILL` exactly when Phase 3 destroys the registry, turning a drain-timeout into a confusing race between manager cleanup and orchestrator kill.

## Fix sketch

1. **Propagate the drain outcome.** Change `wait_for_drain` to return `Result<(), DrainTimeoutError>` with `active_handles` as the payload. `graceful_shutdown` takes the result and decides what to do.

2. **Force-abort vs abort-and-fail.** Let `ShutdownConfig` control the policy:
   ```rust
   pub struct ShutdownConfig {
       pub drain_timeout: Duration,
       /// On drain timeout: `Force` clears the registry (current behavior), `Abort` returns an error
       /// without clearing so the caller can decide.
       pub on_drain_timeout: DrainTimeoutPolicy,
   }
   ```
   Default `Abort` — the loud, blocking, surfaced failure; `Force` is opt-in for \"I know what I'm doing, force the kill\".

3. **Return a result from `graceful_shutdown`.** `-> Result<ShutdownReport, ShutdownError>` where `ShutdownReport` enumerates how many handles were outstanding at each phase. The CLI / supervisor can log it meaningfully.

4. **Lock acquisition on `release_queue_handle.lock().await` is unbounded.** Wrap it in a `tokio::time::timeout` matching the Phase 4 cap so a poisoned/held lock can't stall shutdown forever.

5. **Idempotency guard.** Two concurrent `graceful_shutdown` calls currently both fire `cancel.cancel()`, both call `wait_for_drain`, and the second one's `release_queue_handle.lock().await.take()` returns `None` and silently skips Phase 4. Add a `shutting_down: AtomicBool` CAS guard so the second call returns `Err(AlreadyShuttingDown)` instead of appearing to succeed.

## Related

- `#272` — lost-wakeup race in `wait_for_drain` (already fixed via register-then-check).
- This issue covers the **semantic** gap in `graceful_shutdown` orchestration, not the notify race.

## Severity

**HIGH** — graceful shutdown silently becomes force shutdown on drain timeout, with no caller-visible signal. The whole point of the graceful path is defeated exactly when it matters most (under load, during rolling deploys).

vanyastaff:
PR-A from `docs/plans/2026-04-14-batch3-resource-lifecycle.md`. PR-B (config + credential, #314/#315) shipped as #363.

Four resource lifecycle / concurrency bugs land together because they share Manager error-propagation and DaemonRuntime cancellation surfaces.

## #318 + #323 — DaemonRuntime restart-safe + cancel-responsive backoff
- Per-run `DaemonRun { cancel, handle }` stored under `inner: Mutex<Option<_>>`.
- `start()` builds a fresh child token from `parent_cancel`; `stop()` cancels only the per-run child, leaving `parent_cancel` intact for future starts.
- Stale finished handles are dropped before guarding so start-after-natural-exit succeeds.
- `daemon_loop`'s `restart_backoff` sleep races against the per-run cancel via `biased` `tokio::select!` so `stop()` returns promptly mid-backoff.
- New `is_running()` helper for diagnostics.

## #302 + #322 — graceful_shutdown typed result + recovery gate ticket flow
- `DrainTimeoutPolicy { Abort (default), Force }`, `ShutdownReport`, `ShutdownError` (`DrainTimeout` / `ReleaseQueueTimeout` / `LockTimeout` / `AlreadyShuttingDown`), all `#[non_exhaustive]`.
- `ShutdownConfig` gains `on_drain_timeout` and `release_queue_timeout` fields with builder setters; `Manager::graceful_shutdown` signature changes from `-> ()` to `-> Result<ShutdownReport, ShutdownError>`.
- `AtomicBool shutting_down` CAS guard makes `graceful_shutdown` idempotent.
- `wait_for_drain` now returns `Result<(), DrainTimeoutError { outstanding }>`.
- `check_recovery_gate` replaced by `admit_through_gate` returning `GateAdmission::{Open, Probe(RecoveryTicket)}`; every `acquire_*` path holds a ticket end-to-end and resolves/fails it based on outcome. `trigger_recovery_on_failure` deleted (now structural).

## Test plan
- [x] `cargo clippy -p nebula-resource -- -D warnings`
- [x] `cargo nextest run -p nebula-resource` — 201 passed
- [x] `cargo clippy -p nebula-engine -p nebula-api -- -D warnings` — downstream clean
- [ ] Regression tests for restart-safe daemon and gate herd are deferred to a follow-up; the production fix is verified by existing 201-test suite

Closes #302 #318 #322 #323.

---

> [!NOTE]
> **Medium Risk**
> Changes core resource lifecycle behavior and public API (`Manager::graceful_shutdown` now returns typed errors/reports) plus alters recovery-gate admission and daemon cancellation, which could affect shutdown/retry behavior in downstream callers.
> 
> **Overview**
> Hardens resource lifecycle concurrency in `nebula-resource`.
> 
> `Manager::graceful_shutdown` now returns `Result<ShutdownReport, ShutdownError>` and is CAS-guarded to fail fast on concurrent/second calls. Shutdown behavior on drain timeout is now explicit via `DrainTimeoutPolicy`: default *Abort* returns `DrainTimeout { outstanding }` **without** clearing the registry (handles remain valid), while *Force* clears anyway and reports outstanding handles; release-queue draining is now bounded by configurable `release_queue_timeout` and can error.
> 
> Recovery gating is reworked to be end-to-end ticket owned: acquires CAS-claim a single probe slot up front (`admit_through_gate`) and always settle a `RecoveryTicket` (`resolve`/`fail_transient`/`fail_permanent`), preventing herd stampedes after backoff expiry. `DaemonRuntime` is refactored to be restart-safe with per-run child cancellation tokens and cancel-responsive restart backoff (biased `select!`), plus adds `is_running()`.
> 
> Tests are updated for the new shutdown API and add regressions for drain-timeout policy, AlreadyShuttingDown, and probe-serialization under concurrent acquires; small formatting-only changes occur in config/credential/runtime code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd640a0f45f155899b2b76cb68605c470a2080b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vanyastaff:
Superseded by #397 which landed the resource lifecycle hardening on main. Diff against current main shows only stale code that was rewritten/improved in #397 (e.g. `graceful_shutdown` builder pattern, `GateState` usage). Closing to keep the queue clean — reopen if a specific behavior from this branch needs to be salvaged.

PR #388, opened 2026-04-14
fix(engine): batch 5D credential refresh failure surfaces typed error (#306)

### LINKED ISSUE #306: [MEDIUM] engine NodeTask: proactive credential refresh failure is logged and ignored — broken store causes opaque per-node auth errors
## Summary

`NodeTask::run` calls an optional proactive credential refresh hook before running the action:

```rust
// crates/engine/src/engine.rs ~1517
if let Some(ref refresh_fn) = self.credential_refresh
    && let Err(e) = (refresh_fn)(&self.action_key).await
{
    tracing::warn!(
        node_id = %self.node_id,
        action_key = %self.action_key,
        error = %e,
        \"proactive credential refresh failed; proceeding with potentially stale credential\"
    );
}
```

On any error (credential store down, refresh endpoint unreachable, circuit breaker open, rate limit exceeded, etc.) the code emits a `WARN` and proceeds into `execute_action_versioned`. The action then almost certainly fails with a downstream `401 Unauthorized` or similar, which surfaces as `ActionError::*` with no information that the real cause was a credential-refresh failure a few lines earlier.

## Effect per lifecycle phase

1. **Acquire**: the refresh hook is called per node on every dispatch. No amortization — a workflow with 100 nodes makes 100 refresh calls against the (broken) store.
2. **Use**: the action runs with a stale credential; the downstream HTTP call fails with an opaque auth error.
3. **Partial failure**: no circuit breaker on the refresh path itself — each dispatch re-tries, re-logs, re-fails. Log amplification under a broken store.
4. **Retry**: retries reuse the stale credential (because `ActionResult::Retry` is broken — #296 — but even if it worked, the refresh hook would re-fail the same way).
5. **Cancel**: no cancellation check around the refresh call. A cancel fired during refresh is ignored until the refresh future completes or times out on its own.
6. **Shutdown**: during graceful shutdown, every in-flight node still runs this refresh. If the credential store is also shutting down, refreshes pile up on the dying store.
7. **Restart**: post-restart, the first wave of nodes all hit the refresh hook simultaneously → thundering herd against the credential store (which may still be warming up).
8. **Cleanup**: nothing tracks whether the refresh succeeded, so there's no way to surface \"credential store is broken\" as an execution-level fatal signal. Each node fails independently.

## Why MEDIUM

Not silent data loss (the downstream action's error eventually surfaces), but:

- Root-cause diagnosis for auth errors requires grepping logs for the `WARN` and correlating with action errors.
- Alerting rules on action failures can't distinguish \"credential store broken\" from \"the API we called broke\".
- Under sustained credential-store failure, the engine hammers the store with per-node refresh calls in a tight loop. No backoff, no circuit break.

## Fix sketch

1. **Fail fast when the refresh is mandatory.** The hook should distinguish \"optional nice-to-have refresh\" from \"required: token is expired\". For the \"required\" case, return an error from `NodeTask::run` so the engine's error path runs correctly.

2. **Circuit break the refresh path.** If the last N refresh calls within T seconds all failed, open a circuit and skip proactive refresh entirely until the window closes. Let the action's own call path surface \"credential store unavailable\" once rather than N times.

3. **Cancel-aware refresh.** Wrap the `refresh_fn` call in `tokio::select!` against the cancel token so a shutdown doesn't wait for a dying store.

4. **Typed failure context.** When the action subsequently fails with an auth error, surface a `refresh_preceded_action_failure: true` field on the error so root-cause is visible at the top of the stack, not buried in an earlier `WARN`.

## Severity

**MEDIUM** — observability + diagnosis + thundering-herd on store failure. Correctness-adjacent (the system fails instead of silently using bad data) but operationally bad.

vanyastaff:
## Summary

PR 5D of Batch 5 (engine credential refresh failure handling).

Proactive credential refresh in `NodeTask::run` previously logged a
`WARN` and proceeded with a potentially stale credential. Downstream
auth then failed N times opaquely and operators had no execution-level
signal that the credential store was sick.

This change makes the failure mode typed and routable through
`ErrorStrategy`:

- New `ActionError::CredentialRefreshFailed { action_key, source: Arc<dyn Error + Send + Sync> }` variant — additive (`ActionError` is already `#[non_exhaustive]`).
- Default classification is **retryable** (transient store outages); operators wanting fail-fast can override per action.
- New `EngineError::Action(ActionError)` variant routes the typed error through the existing `handle_node_failure` path with full `Classify` delegation.
- `NodeTask::run` is now cancel-aware: the refresh hook future races against `cancel.cancelled()` via `tokio::select!` so a dying credential store cannot block shutdown.
- On refresh error the node ends up `Failed` with the typed cause in `node_errors`, the action body is **never** invoked, and `ErrorStrategy` decides what happens next.

## Test plan

- [x] `cargo check -p nebula-action -p nebula-engine`
- [x] `cargo clippy -p nebula-action -p nebula-engine --tests -- -D warnings`
- [x] `cargo nextest run -p nebula-engine` — 87 passed, 1 skipped
- [x] `cargo nextest run -p nebula-action` — 445 passed
- [x] New regression test `credential_refresh_failure_surfaces_as_typed_error` asserts:
  1. action handler is never called when refresh fails
  2. workflow result is not a success
  3. `node_errors[n1]` contains both `"credential refresh failed"` and the underlying source message
  4. constructed variant pattern-matches as `EngineError::Action(ActionError::CredentialRefreshFailed { .. })`
- [x] 4 unit tests on `ActionError::CredentialRefreshFailed` covering retryable classification, display, code stability, and Clone

Closes #306

Design doc: `docs/plans/2026-04-14-batch5-misc-lifecycle.md` (PR 5D section)

Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes engine pre-dispatch behavior: proactive credential refresh failures now fail nodes with a typed, retryable error instead of logging and continuing, which can affect workflow outcomes and retry/OnError routing. Adds a new `EngineError` variant and new `ActionError` variant used across crates, so downstream error matching/classification should be revalidated.
> 
> **Overview**
> **Proactive credential refresh failures are now first-class, typed node failures.** The engine’s pre-dispatch `credential_refresh` hook is raced against cancellation and, on error, aborts dispatch and returns `EngineError::Action(ActionError::CredentialRefreshFailed { .. })` rather than logging a warning and continuing.
> 
> **Adds a new retryable `ActionError` variant** (`CredentialRefreshFailed`) with stable `Classify` code/category and a constructor helper, plus **adds `EngineError::Action`** to carry typed action-layer errors through existing classification and `ErrorStrategy` handling. New unit/regression tests assert the action body is not invoked on refresh failure and that the typed error is surfaced in results/events; remaining changes are doc/formatting tidy-ups.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1598f63709a2d89c5bbac5ba13ee8361bbc2447c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vanyastaff:
@copilot resolve the merge conflicts in this pull request

Issue #391, opened 2026-04-14
resource: AcquireOptions intent and tags are never read by topologies


vanyastaff:
## Summary

`AcquireOptions` (`crates/resource/src/options.rs`) documents **`intent`** and **`tags`** as influencing topologies (“may use this to select different pools, apply different timeouts, or skip health checks”).

Within `nebula-resource`, only **`deadline` → `remaining()`** is read on acquire paths (pool, transport, exclusive, etc.). **`intent` and `tags` are never referenced** outside `options.rs` itself (`rg intent crates/resource/src` only hits `options.rs`).

Callers may assume `AcquireIntent::Critical` or routing tags affect behaviour; today they are **no-ops**.

## Suggested direction

- Short term: narrow module-level docs to “reserved for future / engine integration” so public API is not misleading.
- Long term: thread options into topologies (timeouts already use `remaining()`), or drop unused fields until needed.

vanyastaff:
Resolved by PR #399 (merged as f9e97738). Doc-only: `AcquireIntent` (including the `Critical` variant), `AcquireOptions::intent`, and `AcquireOptions::tags` are now explicitly marked as reserved for future engine integration — only `AcquireOptions::deadline` affects acquire behaviour today. The misleading "bypass queues / never throttle" promise has been removed.

Issue #390, opened 2026-04-14
resource: pool Config max_concurrent_creates unused; min_size vs max_size unvalidated


vanyastaff:
## Summary

`crates/resource/src/topology/pooled.rs` `config::Config` exposes **`max_concurrent_creates`**, documented at the workspace level as limiting concurrent instance creation — but **`crates/resource/src/runtime/pool.rs` never references this field** (`rg max_concurrent pool.rs` is empty). Creation paths always go through `create_entry` / acquire with no global create semaphore.

Additionally, there is **no validation** that `min_size <= max_size`. Warmup targets `min_size` and pushes instances into the idle queue **without holding semaphore permits** (documented in `PoolRuntime::warmup`). A mis-set `min_size` much larger than `max_size` can **preallocate far more runtimes than can ever be checked out concurrently**, wasting memory until maintenance/eviction catches up.

## Evidence

- `topology/pooled.rs` — `Config { max_concurrent_creates, min_size, max_size, ... }`
- `runtime/pool.rs` — no `max_concurrent_creates`; warmup uses `target = self.config.min_size as usize` only.

## Suggested direction

- Either implement concurrent-create limiting using `max_concurrent_creates`, or remove/rename the field and fix public docs until implemented.
- Validate `min_size <= max_size` (and sensible bounds for `max_size > 0`) at pool construction or `Manager::register_pooled`, returning `Error::permanent` on violation.

vanyastaff:
Fixed by PR #399 (merged as f9e97738). Two changes: (1) `Manager::register_pooled*` validates `max_size > 0` and `min_size <= max_size` up-front, returning `Error::permanent`, and `PoolRuntime::new` asserts the same invariants as defense-in-depth for direct constructor users. (2) A dedicated `create_semaphore` gates `PoolRuntime::create_entry` to cap concurrent instance creation at `max_concurrent_creates`; both the permit wait and `resource.create` share a single `create_timeout` deadline via `timeout_at`, and the non-blocking `try_acquire` path uses `try_acquire_owned` so its contract is preserved. Covered by `register_pooled_rejects_min_greater_than_max`, `register_pooled_rejects_max_size_zero`, and `pool_create_path_respects_max_concurrent_creates`.

Issue #387, opened 2026-04-14
resource: ResourceStatus / health_check phase never leaves Initializing (ArcSwap never updated)


vanyastaff:
## Summary

`ManagedResource` stores lifecycle metadata in `status: ArcSwap<ResourceStatus>` (see `crates/resource/src/runtime/managed.rs`), initialized in `Manager::register` to `ResourceStatus::new()` — phase **`Initializing`**, embedded `generation` **0**, `last_error` **None**.

Nothing in the crate ever **updates** that `ArcSwap` after registration: there are no `status.store(...)` / phase transitions.

Yet `Manager::health_check` exposes `phase: managed.status().phase` while `generation` comes from `managed.generation()` (the separate `AtomicU64` bumped by `reload_config`). Operators see **phase stuck at `Initializing`** forever and a **correct** hot-reload generation — inconsistent health signal.

## Evidence

- `grep` for `ResourcePhase` / `status` updates under `crates/resource/src`: only `manager.rs` health snapshot + struct definitions.
- `reload_config` bumps `managed.generation` only; does not touch `managed.status`.

## Impact

Any consumer of `ResourceHealthSnapshot` or `ManagedResource::status()` that gates on `phase` or `ResourceStatus::generation` / `last_error` gets **wrong** data in production.

## Suggested direction

Wire phase (at least `Ready` after first successful register path, `Reloading`/`Ready` on reload, etc.) and keep `ResourceStatus.generation` in sync with the atomic **or** stop exposing `phase` until wired; document as "reserved" if intentional stub.

vanyastaff:
Fixed by PR #399 (merged as f9e97738). `ManagedResource` now has `set_phase` / `set_failed` helpers and `AnyManagedResource` exposes `set_phase_erased` so the manager can drive lifecycle phases on registered resources without a typed handle. `register` → `Ready`, `reload_config` → `Reloading → Ready` with the bumped generation, `graceful_shutdown` → `Draining → ShuttingDown` (with `Abort` path restoring `Ready` before returning). Covered by three regression tests in `crates/resource/tests/basic_integration.rs`.

Issue #384, opened 2026-04-14
resource: Exclusive reset runs async while permit already allows next acquire (doc mismatch)


vanyastaff:
## Summary

`Exclusive::reset` in `crates/resource/src/topology/exclusive.rs` is documented as: *"Called when the lease is released, **before the next caller can acquire**."*

In practice, `ExclusiveRuntime::acquire` builds a `ResourceHandle::guarded_with_permit` whose release path only **submits** `reset()` to the shared `ReleaseQueue` (`crates/resource/src/release_queue.rs`) and returns. The binary semaphore permit is dropped **after** the synchronous `on_release` callback finishes (`ResourceHandle` `Drop` in `crates/resource/src/handle.rs` — permit guard dropped at end of `Drop`).

So the **next** exclusive acquire can obtain the permit **while** the previous caller's `reset()` is still **queued or running asynchronously**. That contradicts the trait documentation and can race if `reset` must serialize with the next `acquire`.

## Evidence

- `crates/resource/src/runtime/exclusive.rs` — `acquire` / release callback → `rq.submit(Box::pin(release_exclusive(...)))` only; lease is dropped inside the sync callback (`_returned_lease`), not passed into the async reset path.
- `crates/resource/src/handle.rs` — `Drop for ResourceHandle`: permit retained until after `catch_unwind` around `callback(lease, tainted)` completes (sync), not until release-queue work completes.

## Impact

Resources relying on documented ordering (full reset before any new exclusive use) may see overlapping `reset` vs next use unless they add their own synchronization.

## Suggested direction

Document the actual weaker guarantee **or** block permit release until `reset` completes (e.g. await reset in the drop path — tricky; or use a counter/gate like drain tracker; or run reset synchronously when possible).

vanyastaff:
Fixed by PR #399 (merged as f9e97738). The exclusive `OwnedSemaphorePermit` is now captured inside the submitted release future (`release_exclusive`) and dropped **after** `reset()` resolves, so the next acquirer cannot enter while a previous reset is still running. The trade-off (permit no longer in `HandleInner::Guarded.permit` slot) is documented on `Exclusive::reset` and covered by `exclusive_next_acquire_waits_until_reset_completes` in `crates/resource/tests/basic_integration.rs`.

PR #399, opened 2026-04-14
fix(resource): land + verify all open nebula-resource issues

### LINKED ISSUE #272: [MEDIUM] Manager::wait_for_drain had lost-wakeup race; short-timeout fix applied but drain_timeout still risks stall
## Summary

`crates/resource/src/manager.rs::Manager::wait_for_drain` had the
classic `tokio::sync::Notify` lost-wakeup pattern:

```rust
let drained = tokio::time::timeout(timeout, async {
    loop {
        tracker.1.notified().await;
        if tracker.0.load(AtomicOrdering::Acquire) == 0 {
            return;
        }
    }
}).await;
```

`Notify::notify_waiters()` only wakes waiters that are already
*registered* on the notify queue. `ResourceHandle::Drop` decrements
the counter and, on the `1 → 0` transition, calls
`tracker.1.notify_waiters()`. If that transition happened in the
window between the outer `active == 0` check and the first
`tracker.1.notified().await` poll (lazy registration), the wakeup
was delivered to zero waiters and `wait_for_drain` stalled until the
full `drain_timeout` elapsed — default **30 seconds**.

Under K8s `terminationGracePeriodSeconds = 30s`, this is a guaranteed
`SIGKILL` under any shutdown that closes out the last resource
handle in a short burst — release queue workers killed mid-task,
in-flight cleanup lost.

## Fix applied (2026-04-13 adversarial audit)

`wait_for_drain` now uses the `Notified::enable()` register-then-check
pattern:

```rust
loop {
    let notified = tracker.1.notified();
    tokio::pin!(notified);
    notified.as_mut().enable();  // register BEFORE re-check

    if tracker.0.load(AtomicOrdering::Acquire) == 0 {
        return;  // drain completed during the race window
    }

    notified.await;

    if tracker.0.load(AtomicOrdering::Acquire) == 0 {
        return;
    }
}
```

Two regression tests added in `manager::drain_race_tests`:

- `wait_for_drain_returns_promptly_when_handle_drops` — asserts
  elapsed < 1 s when a handle drops mid-wait.
- `wait_for_drain_catches_drop_via_recheck` — asserts the
  post-`enable()` re-check catches the race.

## Why this issue is still open

Same caveat as #268 (the analogous refresh-path race):

- The window between `try_refresh` / outer load and `enable()` is
  narrowed to a handful of instructions, not eliminated.
- An adversarial scheduler can still interleave `notify_waiters()`
  between mutex release and `enable()`.
- The 30 s default `drain_timeout` still caps the worst case, but
  that is a safety net, not a fix.

Proper closure would move the notification to a primitive that
cannot lose wakeups:

- `tokio::sync::watch` keyed on the active-handle count (version
  number survives missed polls).
- Or `Notified::enable()` called under the same atomic operation
  that decrements the counter (requires restructuring the
  decrement path).

## Discovered

Adversarial audit pass, 2026-04-13
(`.project/audit/adversarial-audit-2026-04-13.md`).

vanyastaff:
## Summary

Resolves the entire open backlog for `nebula-resource` (11 issues).

### Real fixes
- **#382** — scrub stale `TypeId` rows in `Registry::register` when an entry is replaced in place with a different concrete type.
- **#383** — clamp `AcquireResilience::max_attempts` to `>=1` instead of panicking via `expect`.
- **#384** — hold the exclusive `OwnedSemaphorePermit` inside the release future so the next acquirer cannot enter while the previous `reset()` is still running.
- **#387** — drive `ResourceStatus.phase` across `register` (→ `Ready`), `reload_config` (`Reloading → Ready` with generation bumped), and `graceful_shutdown` (`Draining → ShuttingDown`). Adds `set_phase` / `set_failed` helpers and a type-erased `set_phase_erased` on `AnyManagedResource`.
- **#390** — validate `min_size <= max_size && max_size > 0` in `register_pooled` / `register_pooled_with`, and enforce `max_concurrent_creates` via a dedicated `create_semaphore` inside `PoolRuntime::create_entry`.
- **#391** — narrow doc comments for `AcquireOptions::{intent, tags}`: marked as reserved for future engine integration (no behaviour change today).

### Regression coverage for already-fixed issues
- **#272** — verified the existing `wait_for_drain` drain-race tests still pass.
- **#302** — pinned the `DrainTimeoutPolicy::{Abort, Force}` contract with two new tests (`graceful_shutdown_abort_policy_returns_drain_timeout_error`, `graceful_shutdown_force_policy_clears_registry_with_outstanding_count`).
- **#318** — covered `start → stop → start` and `start → natural-exit → start` for `DaemonRuntime`.
- **#322** — asserted `RecoveryGate` single-probe admission under 32 concurrent callers after expiry.
- **#323** — covered `stop()` during `restart_backoff` returning promptly via the `biased select`.

## Test plan

- [x] `cargo nextest run -p nebula-resource` — 217 passed
- [x] `cargo test -p nebula-resource --doc` — 10 passed
- [x] `cargo nextest run --workspace` — 3415 passed
- [x] `cargo clippy --workspace --tests -- -D warnings` — clean
- [x] `cargo deny check` — advisories / bans / licenses / sources ok

Closes #272, #302, #318, #322, #323, #382, #383, #384, #387, #390, #391.

---

> [!NOTE]
> **Medium Risk**
> Touches core resource-manager lifecycle/shutdown paths plus pool/exclusive concurrency controls; behavior changes are guarded by new tests but could impact acquisition/reload/shutdown semantics under load.
> 
> **Overview**
> Closes a batch of `nebula-resource` correctness gaps: clamps misconfigured acquire retry attempts to avoid panics, fixes `Registry::register` replacement to prevent stale `TypeId` lookups, and makes `Manager` actively drive `ResourceStatus.phase` transitions during register/reload/shutdown (including bulk phase updates during graceful shutdown).
> 
> Hardens topology runtimes by holding the exclusive semaphore permit until `reset()` completes, validating pooled config invariants at registration, and enforcing `max_concurrent_creates` via a dedicated create semaphore with a shared timeout budget.
> 
> Adds extensive regression coverage (manager drain-timeout policies, recovery-gate single-probe contention, daemon runtime lifecycle/backoff cancellation, pool create concurrency, exclusive reset ordering) and minor docs/ignore-list tweaks (clarify unused `AcquireOptions` fields; ignore `.worktrees/`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55ed92c283f11aa54a4a3492ef89ba97a2861128. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added validation for pool configurations to reject invalid `min_size` and `max_size` settings.
  * Implemented configurable concurrent resource creation limiting to prevent creation bottlenecks.

* **Bug Fixes**
  * Fixed resilience handling to ensure minimum retry attempts.
  * Improved exclusive resource reset semantics for safer concurrent operations.

* **Improvements**
  * Enhanced resource lifecycle state management during registration, reloading, and shutdown.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -346,3 +346,6 @@ logs/security/
 
 # Claude Code local agent memory (per-developer, not shared)
 .claude/agent-memory-local/
+
+# Git worktrees created by superpowers workflow
+.worktrees/
--- a/crates/resource/src/integration/resilience.rs
+++ b/crates/resource/src/integration/resilience.rs
@@ -92,27 +92,32 @@ impl AcquireResilience {
     /// Convert to a [`RetryConfig`] for use with [`nebula_resilience::retry_with`].
     ///
     /// Maps exponential backoff (2× multiplier) and optional wall-clock
-    /// timeout budget. When no retry config is set, returns a single-attempt
-    /// config with optional timeout budget.
-    pub(crate) fn to_retry_config<E: 'static>(&self) -> RetryConfig<E> {
+    /// timeout budget. Returns `None` if [`RetryConfig::new`] rejects the
+    /// (clamped, `>= 1`) attempt count — today that path is unreachable,
+    /// but a future version of `nebula-resilience` might add new
+    /// validation and we prefer "skip retry" over "panic the manager".
+    /// Callers that receive `None` should run the operation without
+    /// retry (the existing `None` branch in [`execute_with_resilience`]).
+    pub(crate) fn to_retry_config<E: 'static>(&self) -> Option<RetryConfig<E>> {
+        // #383: clamp a user-supplied `max_attempts: 0` to 1 so we
+        // degrade to a single attempt instead of panicking. Clamping is
+        // documented on `AcquireRetryConfig::max_attempts`.
         let max_attempts = self.retry.as_ref().map_or(1, |r| r.max_attempts.max(1));
-        let cfg = RetryConfig::new(max_attempts).expect("max_attempts clamped to >=1 above");
+        let mut cfg = RetryConfig::new(max_attempts).ok()?;
 
-        let cfg = if let Some(ref retry) = self.retry {
-            cfg.backoff(BackoffConfig::Exponential {
+        if let Some(ref retry) = self.retry {
+            cfg = cfg.backoff(BackoffConfig::Exponential {
                 base: retry.initial_backoff,
                 multiplier: 2.0,
                 max: retry.max_backoff,
-            })
-        } else {
-            cfg
-        };
+            });
+        }
 
         if let Some(timeout) = self.timeout {
-            cfg.total_budget(timeout)
-        } else {
-            cfg
+            cfg = cfg.total_budget(timeout);
         }
+
+        Some(cfg)
     }
 }
 
--- a/crates/resource/src/integration/resilience.rs
+++ b/crates/resource/src/integration/resilience.rs
@@ -36,7 +36,7 @@ pub struct AcquireRetryConfig {
     /// Maximum total number of attempts (including the initial try).
     ///
     /// For example, `max_attempts: 3` means 1 initial try + 2 retries.
-    /// Values below 1 are clamped to 1 by the manager.
+    /// Values below 1 are clamped to 1 at conversion time.
     pub max_attempts: u32,
     /// Initial backoff duration before the first retry.
     pub initial_backoff: Duration,
@@ -95,9 +95,6 @@ impl AcquireResilience {
     /// timeout budget. When no retry config is set, returns a single-attempt
     /// config with optional timeout budget.
     pub(crate) fn to_retry_config<E: 'static>(&self) -> RetryConfig<E> {
-        // #383: a misconfigured `max_attempts: 0` from a config file used to
-        // panic here via `expect`. Clamp to `1` so we degrade to a single
-        // attempt rather than killing the process during acquire.
         let max_attempts = self.retry.as_ref().map_or(1, |r| r.max_attempts.max(1));
         let cfg = RetryConfig::new(max_attempts).expect("max_attempts clamped to >=1 above");
 
--- a/crates/resource/src/integration/resilience.rs
+++ b/crates/resource/src/integration/resilience.rs
@@ -36,6 +36,7 @@ pub struct AcquireRetryConfig {
     /// Maximum total number of attempts (including the initial try).
     ///
     /// For example, `max_attempts: 3` means 1 initial try + 2 retries.
+    /// Values below 1 are clamped to 1 by the manager.
     pub max_attempts: u32,
     /// Initial backoff duration before the first retry.
     pub initial_backoff: Duration,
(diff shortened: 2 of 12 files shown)


vanyastaff:
## Review response — round-3 fixes (commit `55ed92c2`)

### Resolved

**Cursor Bugbot**
- ✅ **Lock-ordering reversal in `Registry::register`** (`5104e517`) — `entries` guard is dropped before touching `type_index`; strict one-way order `entries → type_index`. Re-comment in `23c994cd` rewords the note to get past the `typos` check.
- ✅ **`create_semaphore.acquire_owned` unbounded** (`23c994cd`) — whole `create_entry` path now shares a single `create_timeout` deadline via `timeout_at`.
- ✅ **Abort path leaves resources stuck in `Draining`** (`55ed92c2`) — `graceful_shutdown` now restores `Ready` before returning `DrainTimeout` so `health_check` and any retry start from a sane state.

**CodeRabbit**
- ✅ **`type_index` cleanup breaks other scope with same type** (`55ed92c2`) — `Registry::register` scans the remaining entries under the key and only scrubs the stale `TypeId` row when no other scope still uses that concrete type. Added `register_replace_preserves_type_id_still_used_by_another_scope` regression test.
- ✅ **`try_acquire` non-blocking broken by create-semaphore wait** (`55ed92c2`) — `create_entry` now takes a `non_blocking` flag; `try_acquire` sets it to `true` and returns `Backpressure` immediately if the create semaphore is saturated.
- ✅ **Gate admission test was sequential** (`5104e517`) — rewritten with `tokio::spawn` + `Barrier::new(32)` so the CAS path is actually contended.
- ✅ **`AcquireIntent::Critical` doc still promised bypass** (`55ed92c2`) — variant doc now matches the enum-level `#391` "reserved for future integration" note.

**Copilot**
- ✅ **`.expect` in `to_retry_config`** (`5104e517`) — signature is now `Option<RetryConfig>`; `execute_with_resilience` falls through to a single un-retried attempt if the conversion ever returns `None` (which is currently unreachable).
- ✅ **`PoolRuntime::new` accepts invalid config** (`5104e517`) — defense-in-depth: `assert!` on `max_size > 0` and `min_size <= max_size` inside the public constructor so direct callers get a loud failure instead of a silent deadlock.
- ✅ **Misleading `reload_config` comment about generation** (`5104e517`) — rewritten to match what `health_check` actually reads (the atomic).

### Push-back

**Cursor Bugbot — `exclusive.rs` permit inside closure loses panic-safety** (Low)
Tracked and intentional. The old `HandleInner::Guarded { permit, ... }` pattern drops the permit at the *end* of the handle's `Drop` scope — i.e. **before** the async `reset()` actually runs on the release-queue worker, so it does not give the `#384` guarantee. To keep the permit alive until `reset()` resolves, it has to live inside the submitted future. The release callback body is `m.record_release()` (metric counter increment) and `rq.submit(...)` (channel `try_send` with fallback) — neither is a realistic panic source, so the trade-off is low risk for a Medium-severity correctness win.

**Copilot — adding required methods to `AnyManagedResource` is a breaking change**
Tracked and accepted. `nebula-resource` is alpha-stage per `CLAUDE.md` (*"Breaking changes are acceptable when they remove bad APIs or reduce complexity"*). The only real `impl` is `ManagedResource<R>`; no downstream crate exposes a custom implementation today. `managed_type_id` has no sensible default — returning `TypeId::of::<()>()` or similar would silently break the `#382` stale-row scrub. Leaving the methods required so any future external implementors get a loud compile error is the intended behaviour.

---

Local gate on `55ed92c2`: 218 tests passing, `cargo clippy --workspace --tests -- -D warnings` clean. CodSpeed CI failure is pre-existing infra (`No benchmarks found for walltime mode`) unrelated to this PR.

Issue #348, opened 2026-04-14
[MEDIUM] InMemoryPendingStore get() retains expired entries (stale-row memory growth)


vanyastaff:
## Summary
`InMemoryPendingStore::get()` returns `Expired` for stale entries but does not evict them from the backing map. Expired rows are only removed on `consume()`.

## Impact
- In long-running dev/test environments, repeated `put` + expired `get` patterns can grow memory without bound.
- Token namespace keeps stale entries indefinitely, increasing lookup overhead and diagnostic noise.

## Failure Scenario
1. Pending tokens are created with short TTL.
2. Caller periodically checks state via `get()` after expiry (without `consume()`).
3. Each call returns `PendingStoreError::Expired`, but entry remains in map.
4. Map accumulates expired records over time.

## Evidence
- `crates/credential/src/pending_store_memory.rs`:
  - `get()`:
    - checks `Utc::now() > entry.expires_at` and returns `Expired`.
    - does not remove key from `entries`.
  - `consume()`:
    - on expiry path explicitly calls `entries.remove(token.as_str())`.

## Why Tests/Lints Missed It
- Existing tests assert `Expired` behavior, but not post-condition on storage size/eviction after `get()`.
- Static analysis does not flag lifecycle asymmetry between `get` and `consume`.

## Proposed Fix
- Evict expired entries on `get()` as well (safe because expiration is deterministic).
- Add regression test proving map size does not grow when repeatedly reading expired tokens.

## Acceptance Criteria
- Expired entry is removed regardless of whether caller uses `get()` or `consume()`.
- Repeated expired `get()` calls do not retain stale entries.

## Related Files
- `crates/credential/src/pending_store_memory.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

vanyastaff:
Fixed in main as `c6df5a32` — "fix(credential): evict expired pending rows on get (#348)". Issue stayed open because the squash-merge subject did not contain the GitHub-recognized auto-close keyword pattern. Closing manually after verifying tests pass.

Issue #282, opened 2026-04-14
[MEDIUM] Lazy re-encryption returns StoredCredential with stale version after CAS write — phantom conflicts


vanyastaff:
## Summary

`EncryptionLayer` lazy re-encryption path performs a CAS write to persist the re-encrypted blob but returns the `StoredCredential` struct to the caller with the **pre-write** `version` field. The caller then attempts its own CAS update with that stale version and gets a phantom conflict against the fresh row the encryption layer just wrote.

## Location

`crates/credential/src/layer/encryption.rs:131`

## Failure sequence

1. Caller calls `resolver.resolve(cred_id)`.
2. Resolver pulls row at `version = N`.
3. Encryption layer detects an old `key_id` on the blob, re-encrypts, writes back at `version = N + 1`.
4. Encryption layer returns the decrypted state to the resolver, but with `version = N` on the struct.
5. Resolver finishes its refresh, attempts CAS on `version = N`, hits the freshly-incremented row → CAS fails with `VersionMismatch`.
6. Resolver retries, observes `version = N + 1`, sees its own write missing, retries again — observability becomes \"random refresh conflicts\".

The re-encryption is invisible to the audit log; the conflict is visible — operators end up chasing ghost contention in the wrong layer.

## Fix sketch

After the re-encrypt CAS write, update the in-memory `StoredCredential.version` to the value the CAS write returned before handing it back to the caller.

## Severity

MEDIUM — not data loss, but a correctness bug in an invariant (\"returned version matches persisted version\") that callers assume holds.

## Related

Finding `BL-C6-17` from `.project/audit/2026-04-14-deep-review.md`.

Issue #281, opened 2026-04-14
[MEDIUM] EncryptionLayer single-key constructor aliases empty string key_id — legacy records silently decrypt


vanyastaff:
## Summary

`EncryptionLayer::new(key)` (the single-key convenience constructor) registers the key under both its own `key_id` and under the empty string `""`, and `encrypt()` produces envelopes with `key_id: ""`. This breaks the \"no legacy fallback\" migration story: records persisted by any prior single-key version of the layer are silently decrypted with the current key, and conversely current encrypts are indistinguishable from legacy envelopes.

## Location

`crates/credential/src/layer/encryption.rs:67`

## Risks

1. A caller expecting rotation (\"old records encrypted with key v1 should error out after retiring v1\") will instead see them silently decrypt with v2.
2. Audit logs that lean on `key_id` for provenance become unreliable — every \"`\"\"`\" entry is really \"whichever key was current at encrypt time\".
3. A multi-key deployment mistakenly touching the single-key constructor for any reason pollutes the envelope stream with empty `key_id`, mixing cohorts permanently.

## Fix sketch

Remove the empty-string alias from the map. Force the single-key constructor to use a real `key_id` (caller-provided, or derived from the key hash). Migrate any existing deployments via a one-time re-encrypt job that reads `key_id: \"\"` as \"legacy, re-encrypt with current key\".

## Severity

MEDIUM — does not leak plaintext, but breaks encryption-layer invariants that the system later relies on for auditing and key rotation safety.

## Related

Finding `BL-C4-10` from `.project/audit/2026-04-14-deep-review.md`.

Issue #268, opened 2026-04-14
[MEDIUM] Refresh waiter lost-wakeup race not fully closed by Notified::enable()


vanyastaff:
## Summary

`CredentialResolver::resolve_and_refresh`'s `RefreshAttempt::Waiter`
arm has a narrow but real race with `Notify::notify_waiters()`:

1. Waiter `T2` takes the coordinator mutex, gets `Waiter(Arc<Notify>)`,
   releases the mutex.
2. *Before* `T2` reaches `tokio::pin!(notified); notified.as_mut().enable();`,
   winner `T1` finishes, fires scopeguard → `complete()` +
   `notify_waiters()`.
3. `T2`'s later `enable()` runs against a `Notify` that has no pending
   permit from `notify_waiters()` (that function does not buffer), so
   the wakeup is lost.

In practice the window is a handful of instructions — sub-microsecond —
so the race is extremely unlikely under a non-cached HTTP refresh. But
under a cached / stub / sub-ms refresh, and under adversarial scheduling,
it can still fire.

The audit pass on 2026-04-13 already:

- narrowed the window by adding `tokio::pin!` + `.as_mut().enable()`
  before any await (was: lazy first-poll registration)
- shortened the waiter-side timeout from 60 s → 5 s
- kept the post-wait `resolve::<C>()` re-read as the race-recovery
  safety net

so a lost wakeup now costs **up to 5 s** of latency, not up to 60 s, and
always recovers to the right answer. This issue tracks closing the
residual race properly.

## Impact

- **Medium.** Even at 5 s, a burst of refreshes at the top of the hour
  (many credentials expiring simultaneously) can still push a fraction
  of waiters through the worst-case path, spending 5 s idle. Under
  pathological scheduler behaviour, ALL waiters could hit it — 5 s
  per credential per worker is still an SLO breach.

## Proper fix options

**Option A — watch channel with version counter.**

Replace `HashMap<String, Arc<Notify>>` with
`HashMap<String, tokio::sync::watch::Sender<u64>>`. Winner bumps the
version on completion; waiters subscribe on the sender side *inside*
the locked `try_refresh` call and the `Receiver` they get back carries
the version at time of subscription, so a post-subscribe version bump
is seen unambiguously. No wakeup can be "missed."

**Option B — return a pre-registered `Notified`.**

Currently `Waiter(Arc<Notify>)` leaks the race by forcing the caller
to do registration after the mutex is released. If `try_refresh`
[...]

PR #407, opened 2026-04-15
fix(credential): OAuth2 safe errors, refresh validation, bounded circuit breakers

### LINKED ISSUE #274: [MEDIUM] OAuth2 refresh silently ignores response missing access_token — stale token + fresh expiry
## Summary

`oauth2_flow::refresh_token` treats a 200 OK refresh response that is missing the `access_token` field as a successful refresh. The stale access token is kept in `OAuth2State`, but other fields (`expires_at`, `scopes`, `refresh_token`) may be updated from the response, masking the staleness.

## Location

`crates/credential/src/credentials/oauth2_flow.rs:501` (`update_state_from_token_response`) called from `refresh_token` at `:422`.

## Failure path

```rust
pub(crate) async fn refresh_token(state: &mut OAuth2State, config: &OAuth2Config) -> Result<(), CredentialError> {
    // ... HTTP POST ...
    let body: Value = parse_token_response(resp).await?;  // OK on 2xx
    update_state_from_token_response(state, &body);       // silently skips missing access_token
    Ok(())
}

fn update_state_from_token_response(state: &mut OAuth2State, body: &Value) {
    if let Some(token) = body.get("access_token").and_then(Value::as_str) {
        state.access_token = SecretString::new(token);
    }
    if let Some(secs) = body.get("expires_in").and_then(Value::as_u64) {
        state.expires_at = Some(Utc::now() + chrono::Duration::seconds(secs as i64));
    }
    // ...
}
```

If a provider returns `{"expires_in": 3600}` without `access_token` (malformed, bug in provider, or 200 with an error-like body), then:

- `state.access_token` remains the stale pre-refresh value
- `state.expires_at` gets bumped forward an hour
- `refresh_token()` returns `Ok(())`
- `CredentialResolver` believes the token is valid and hands it out
- Downstream HTTP calls fail with 401 an hour later, with no sign of the original refresh failure

RFC 6749 §5.1 REQUIRES `access_token` in any successful token response — a 200 without it is always provider misbehavior, but the client side must surface it as an error rather than silently succeed.

## Why tests missed it

`state_from_token_response` (the *initial* exchange path) does error on missing `access_token`. `update_state_from_token_response` (the refresh path) was written to \"only overwrite fields present\", which was correct for *optional* fields (`refresh_token`, `scope`) but wrongly inherited by the *required* `access_token`.

Tests cover the happy path (`refresh_token_updates_fields`) and preservation of refresh token across refresh (`update_state_preserves_existing_refresh_token`), but no test constructs a 200 response without `access_token`.

## Fix sketch

Have `update_state_from_token_response` return `Result<(), CredentialError>` and error out when `access_token` is absent:

```rust
fn update_state_from_token_response(state: &mut OAuth2State, body: &Value) -> Result<(), CredentialError> {
    let Some(token) = body.get("access_token").and_then(Value::as_str) else {
        return Err(provider_error(
            "refresh response missing required 'access_token' field".into(),
        ));
    };
    state.access_token = SecretString::new(token);
    // ... rest unchanged ...
    Ok(())
}
```

Surface `CredentialError::Provider` so the resolver's circuit breaker + retry path observes the failure instead of a stale OK.

Regression test: POST handler that returns 200 `{\"expires_in\": 3600}`, assert `refresh_token` returns `Err(Provider(_))`.

## Severity

MEDIUM — correctness + hidden staleness. Not exploitable by an attacker without a cooperating (or broken) IdP, but masks incidents and makes root-causing 401s downstream far harder.

### LINKED ISSUE #277: [MEDIUM] OAuth2 token endpoint error body interpolated into error message — provider-side secret echo leaks into logs
## Summary

`parse_token_response` in `crates/credential/src/credentials/oauth2_flow.rs:431` takes a non-2xx HTTP response from the OAuth2 token endpoint and interpolates **the entire response body** into a `CredentialError::Provider` message:

```rust
async fn parse_token_response(resp: reqwest::Response) -> Result<Value, CredentialError> {
    let status = resp.status();
    if !status.is_success() {
        let body_text = resp.text().await.unwrap_or_default();
        return Err(provider_error(format!(
            \"token endpoint returned {status}: {body_text}\"
        )));
    }
    // ...
}
```

## Why this matters

Some OAuth2 providers echo caller-supplied fields back in `400 Bad Request` / `401 Unauthorized` envelopes. Observed in the wild:

- GitLab: returns `{\"error_description\": \"...\", \"request\": {\"client_id\": \"...\", \"client_secret\": \"...\"}}` in certain dev/staging builds
- Azure AD: echoes the submitted `client_id` in several error bodies
- Some Keycloak realms echo `refresh_token` when they think the value is malformed

`CredentialError::Provider(String)` gets:

1. Rendered into `Display` output and logged by the resolver / audit layer
2. Propagated up through `CredentialResolver::resolve` and into `ActionError`
3. Captured by the audit sink, which the credential audit contract says must **never** contain credential data

Effectively: any token-endpoint misbehavior that echoes credentials leaks those credentials into our audit log and structured tracing. The audit layer's \"never log credential data\" contract is broken by a single `format!` call.

## Location

`crates/credential/src/credentials/oauth2_flow.rs:431-438`

Same pattern is used from:
- `exchange_client_credentials` → `parse_token_response`
- `exchange_authorization_code` → `parse_token_response`
- `refresh_token` → `parse_token_response`
- `poll_device_code` does its own `body.get(\"error\")` parse and is safe

## Fix sketch

Do not include the raw body in the error. Pull out only known-safe fields (`error`, `error_description`, `error_uri`) via `serde_json::from_str`, and log only those:

```rust
async fn parse_token_response(resp: reqwest::Response) -> Result<Value, CredentialError> {
    let status = resp.status();
    if !status.is_success() {
        // Only extract RFC 6749 §5.2 standard error fields — never the raw body.
        let body_text = resp.text().await.unwrap_or_default();
        let safe = serde_json::from_str::<serde_json::Value>(&body_text)
            .ok()
            .as_ref()
            .and_then(|v| v.get(\"error\").and_then(|e| e.as_str()).map(str::to_owned))
            .unwrap_or_else(|| \"<no error code>\".into());
        return Err(provider_error(format!(
            \"token endpoint returned {status} ({safe})\"
        )));
    }
    // ...
}
```

If richer diagnostics are needed, surface the HTTP status + RFC 6749 §5.2 `error` code only.

## Regression test

`mockito`-style test: mock token endpoint returns `400` with `{\"error\":\"invalid_client\",\"client_secret\":\"hunter2\"}`. Assert the resulting `CredentialError::Provider` message does not contain `hunter2`.

## Severity

MEDIUM — not a self-inflicted leak, but passively leaks anything a provider chooses to echo. Audit-contract violator.

## Related

Finding `BL-C4-07` from `.project/audit/2026-04-14-deep-review.md`.

### LINKED ISSUE #278: [MEDIUM] RefreshCoordinator::circuit_breakers grows without bound — slow memory leak across credential churn
## Summary

`RefreshCoordinator::circuit_breakers: Mutex<HashMap<String, Arc<CircuitBreaker>>>` only removes entries on `record_success`. Any credential that has **ever** failed refresh and has not yet had a successful refresh leaves an `Arc<CircuitBreaker>` wedged in the map for the entire process lifetime.

## Location

`crates/credential/src/refresh.rs:147` (field definition) · `:153` (`get_or_create_cb` inserts) · `:227` (`record_failure` inserts, never removes) · `:235` (`record_success` is the only remover)

```rust
pub fn record_failure(&self, credential_id: &str) {
    self.get_or_create_cb(credential_id)
        .record_outcome(Outcome::Failure);
}

pub fn record_success(&self, credential_id: &str) {
    // Remove the CB entirely on success — full reset.
    self.circuit_breakers.lock().remove(credential_id);
}
```

## Failure scenarios

1. **Credential deletion.** An operator deletes a credential that was failing refresh. The CB entry is never swept.
2. **Expired credential with no refresh_token.** Refresh returns `ReauthRequired`, which counts as a failure on the CB path but never a success. Entry stays until process restart.
3. **Long-lived process + credential churn.** Every new credential that ever fails a single refresh leaves a `CircuitBreaker` behind. Memory grows roughly linearly with \"distinct credentials that ever failed a refresh in this process\".

Each `CircuitBreaker` carries a sliding-window buffer + atomics — small, but unbounded × small = eventually large.

## Attacker amplification

If `credential_id` strings are ever attacker-shapable (e.g. the ID includes a user-provided prefix, or a misconfigured system lets external input flow into the ID), a bad actor can allocate arbitrary CB entries by driving repeated refresh failures. Memory DoS on a long-running daemon.

Current `CredentialId::new()` is a UUID so attacker-shaping is not a present concern, but the unbounded growth is still a correctness smell and is trivially triggered by legitimate credential churn.

## Fix options

1. **LRU cap.** Replace the raw `HashMap` with a bounded LRU (e.g. `lru::LruCache`) sized at N × expected-active-credentials.
2. **TTL-based eviction.** Timestamp each entry on last-updated and evict entries that have been in their current state for > `reset_timeout` × 2. Hook eviction into `get_or_create_cb` so normal traffic sweeps stale state.
3. **Explicit removal on credential deletion.** Expose a `forget(credential_id)` method and call it from `CredentialManager::delete`. Combined with (2) this covers both failure modes.

Option (2) is lowest-surface and self-healing.

Also: `in_flight: Mutex<HashMap<String, Arc<Notify>>>` has the same shape. `try_refresh` inserts, `complete` removes — `complete` is the only remover, and if a refresh panics *and* the scopeguard is somehow bypassed, the entry leaks. Lower risk but worth the same treatment.

## Related

Finding `BL-C2-15` from `.project/audit/2026-04-14-deep-review.md`.

## Severity

MEDIUM — not immediately exploitable today (UUID IDs), but the unbounded-growth smell is latent and becomes real the moment IDs are derived from anything external.

vanyastaff:
## Summary

Addresses three open **nebula-credential** issues from the audit.

### Issues closed

- [#277](https://github.com/vanyastaff/nebula/issues/277): Non-2xx token responses no longer interpolate the raw body into errors. Only RFC 6749 error fields are summarized (error code, truncated error_description, error_uri).
- [#274](https://github.com/vanyastaff/nebula/issues/274): Refresh responses must include `access_token`; otherwise we return `CredentialError::Provider` and do not update expiry while leaving a stale token.
- [#278](https://github.com/vanyastaff/nebula/issues/278): Per-credential circuit breakers are stored in an LRU capped at 4096 entries (dependency: `lru` 0.16.3 for advisory alignment with the rest of the lockfile).

### Tests added

- Safe OAuth2 error summary does not echo provider-echoed secrets in sibling JSON keys.
- Missing `access_token` on refresh body yields error and preserves prior token.
- Circuit breaker map stays bounded under churn.

### Follow-ups (still open)

Other `nebula-credential` issues not covered here: #265, #268, #281, #282, #348.

Closes #274
Closes #277
Closes #278

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth2 token refresh/error handling and refresh coordination logic; regressions could affect credential refresh reliability, though changes add validation, safer logging, and tests.
> 
> **Overview**
> Prevents leaking provider error bodies in OAuth2 token endpoint failures by replacing raw-body interpolation with a log-safe RFC 6749 error summary (including truncated `error_description`).
> 
> Tightens refresh handling by requiring `access_token` in successful refresh responses and returning `CredentialError::Provider` otherwise, avoiding updates that would keep a stale token.
> 
> Bounds refresh circuit breaker tracking by switching to an LRU cache capped at 4096 entries (adds `lru` dependency) and adds targeted tests for secret-safe error summaries, missing-token refresh responses, and bounded breaker growth.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fc54afd96ff8a84cc9b9f50c6e40e27b1684ccd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security in OAuth2 error handling to prevent sensitive authentication information from being inadvertently logged
  * Enhanced validation logic to ensure access tokens are present in token refresh responses
  * Fixed potential unbounded memory growth in credential refresh mechanisms by implementing bounded tracking for circuit breakers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3845,6 +3845,7 @@ dependencies = [
  "chrono",
  "futures",
  "humantime-serde",
+ "lru",
  "moka",
  "nebula-core",
  "nebula-credential-macros",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ tracing-subscriber = { version = "0.3.23", default-features = false }
 
 # Concurrency and synchronization
 dashmap = { version = "6.1.0", default-features = false }
+lru = { version = "0.16.3", default-features = false }
 parking_lot = { version = "0.12.5", default-features = false }
 moka = { version = "0.12", default-features = false }
 once_cell = "1.21"
--- a/crates/credential/Cargo.toml
+++ b/crates/credential/Cargo.toml
@@ -57,6 +57,7 @@ tracing = { workspace = true }
 humantime-serde = "1.1"
 
 # Caching
+lru = { workspace = true }
 moka = { workspace = true, features = ["future"] }
 
 # Synchronization
--- a/crates/credential/src/credentials/oauth2_flow.rs
+++ b/crates/credential/src/credentials/oauth2_flow.rs
@@ -443,9 +443,10 @@ fn oauth_token_error_summary(body_text: &str) -> String {
     let mut out = error.to_owned();
     if let Some(desc) = value.get("error_description").and_then(Value::as_str) {
         out.push_str(": ");
-        let truncated: String = desc.chars().take(256).collect();
-        out.push_str(&truncated);
-        if desc.chars().count() > 256 {
+        // Bound work: at most 257 scalar values, never scan the full provider string.
+        let prefix: Vec<char> = desc.chars().take(257).collect();
+        out.extend(prefix.iter().take(256).copied());
+        if prefix.len() > 256 {
             out.push('…');
         }
     }
--- a/crates/credential/src/credentials/oauth2_flow.rs
+++ b/crates/credential/src/credentials/oauth2_flow.rs
@@ -422,11 +422,40 @@ pub(crate) async fn refresh_token(
         .map_err(|e| provider_error(format!("refresh token request failed: {e}")))?;
 
     let body: Value = parse_token_response(resp).await?;
-    update_state_from_token_response(state, &body);
+    update_state_from_token_response(state, &body)?;
     Ok(())
 }
 
 // ── Internal helpers ────────────────────────────────────────────────────
+
+/// Builds a log-safe summary for a non-2xx OAuth2 token endpoint body.
+///
+/// Never interpolates the raw response body: some providers echo submitted
+/// secrets in error JSON. Only RFC 6749 §5.2 fields are included, with
+/// `error_description` truncated (see GitHub issue #277).
+fn oauth_token_error_summary(body_text: &str) -> String {
+    let Ok(value) = serde_json::from_str::<Value>(body_text) else {
+        return "<non-json body>".to_owned();
+    };
+    let Some(error) = value.get("error").and_then(Value::as_str) else {
+        return "<no error code>".to_owned();
+    };
+    let mut out = error.to_owned();
+    if let Some(desc) = value.get("error_description").and_then(Value::as_str) {
+        out.push_str(": ");
+        let truncated: String = desc.chars().take(256).collect();
+        out.push_str(&truncated);
+        if desc.chars().count() > 256 {
+            out.push('…');
+        }
+    }
+    if let Some(uri) = value.get("error_uri").and_then(Value::as_str) {
+        out.push_str(" (error_uri=");
+        out.push_str(uri);
+        out.push(')');
+    }
+    out
+}
 
 /// Parse an HTTP response as a JSON token response.
 ///
@@ -435,8 +464,9 @@ async fn parse_token_response(resp: reqwest::Response) -> Result<Value, Credenti
     let status = resp.status();
     if !status.is_success() {
         let body_text = resp.text().await.unwrap_or_default();
+        let summary = oauth_token_error_summary(&body_text);
         return Err(provider_error(format!(
-            "token endpoint returned {status}: {body_text}"
+            "token endpoint returned {status}: {summary}"
         )));
     }
     resp.json::<Value>()
@@ -499,12 +529,22 @@ fn state_from_token_response(
 
 /// Update an existing [`OAuth2State`] from a refresh token response.
 ///
-/// Only overwrites fields present in the response. A missing `refresh_token`
+/// RFC 6749 §5.1 requires `access_token` in a successful token response; a 2xx
+/// body without it is treated as an error so we do not bump `expires_at` while
+/// leaving a stale access token (GitHub issue #274).
+///
+/// Other fields are only overwritten when present. A missing `refresh_token`
 /// preserves the existing one (per RFC 6749 Section 6).
-fn update_state_from_token_response(state: &mut OAuth2State, body: &Value) {
-    if let Some(token) = body.get("access_token").and_then(Value::as_str) {
-        state.access_token = SecretString::new(token);
-    }
+fn update_state_from_token_response(
+    state: &mut OAuth2State,
+    body: &Value,
+) -> Result<(), CredentialError> {
+    let Some(token) = body.get("access_token").and_then(Value::as_str) else {
+        return Err(provider_error(
+            "refresh response missing required 'access_token' field".into(),
+        ));
+    };
+    state.access_token = SecretString::new(token);
     if let Some(tt) = body.get("token_type").and_then(Value::as_str) {
         state.token_type = tt.to_owned();
     }
@@ -517,6 +557,7 @@ fn update_state_from_token_response(state: &mut OAuth2State, body: &Value) {
     if let Some(scope) = body.get("scope").and_then(Value::as_str) {
         state.scopes = scope.split_whitespace().map(str::to_owned).collect();
     }
+    Ok(())
 }
 
 /// Build a `CredentialError::Provider` from an HTTP/network-related message.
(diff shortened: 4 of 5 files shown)


vanyastaff:
Addressed the Copilot review comments on this PR:

1. **`circuit_breaker_tracking_is_bounded`** — the assertion now uses `MAX_TRACKED_CIRCUIT_BREAKERS.get()` instead of a literal `4096`, so it stays tied to the LRU cap constant.

2. **`oauth_token_error_summary`** — `error_description` is processed with at most 257 scalar iterations (`take(257)` + `prefix.len() > 256` for the ellipsis). We no longer call `chars().count()` on the full provider-controlled string.

3. **`MAX_TRACKED_CIRCUIT_BREAKERS`** — Copilot suggested avoiding `unwrap()` via `unsafe` `new_unchecked`, but this crate uses `#![forbid(unsafe_code)]`. The cap is now `NonZeroUsize::new(MAX_TRACKED_CIRCUIT_BREAKERS_CAP).unwrap()` in a `const` (compile-time only, with a short doc note). Added `MAX_TRACKED_CIRCUIT_BREAKERS_CAP` for the raw `usize` value.

4. **`CircuitBreaker::new(...).expect(...)`** — replaced with `CircuitBreaker::new(config)` and, on `Err`, `tracing::error!` plus fallback to `CircuitBreaker::new(CircuitBreakerConfig::default())`. If that still fails, `unreachable!` documents the nebula-resilience invariant (should never happen in practice).

Changes are pushed to this branch.

PR #428, opened 2026-04-18
fix(engine): batch 2 execution-state correctness (#299, #300, #301, #311, #321)

### LINKED ISSUE #299: [HIGH] check_and_apply_idempotency reconstructs ActionResult as Success — Branch/Route/MultiOutput semantics lost on resume

### LINKED ISSUE #300: [MEDIUM] engine: spawn_node's chained transition_to calls swallow invalid transitions — node runs in inconsistent state

### LINKED ISSUE #301: [MEDIUM] engine run_frontier: panicked NodeTask reports NodeId::new() — cannot identify which node panicked

### LINKED ISSUE #311: [HIGH] resume_execution drops original workflow input (replays with Null)

### LINKED ISSUE #321: [HIGH] Engine setup-failure path skips checkpoint, losing state on crash before final persist

vanyastaff:
## Summary

Resurrection of [PR #346](https://github.com/vanyastaff/nebula/pull/346) ("batch 2 execution-state correctness"), which was approved in spirit, **closed without merging**, and never came back. Subsequent PRs (e.g. `6c12a127` / batch 5C) referenced #346 as if it had landed — spot-checking main at `2b205abf` confirmed the phantom-fix belief and that the five tracked bugs are genuinely still real. Per tech-lead's 2026-04-18 directive: "Resurrect, don't re-derive."

Single commit cherry-picked from `0f29b62b` and adapted for current main — primarily the `NodeId → NodeKey` rename from [#412](https://github.com/vanyastaff/nebula/pull/412).

## Five fixes

- **#321 — setup-failure path now checkpoints.** `run_frontier`'s setup-failure branch calls `checkpoint_node` + emits `NodeFailed`, matching the runtime-failure ordering. Previously in-memory state moved to `Failed` while the persisted row still said `Pending`, producing ghost-pending nodes visible to CAS readers and `resume_execution`.
- **#300 — typed node transitions.** `spawn_node` drives state through a new `NodeExecutionState::start_attempt` helper modelling the legal paths (`Pending → Ready → Running`, `Failed → Retrying → Running`, `Retrying → Running`). `ExecutionState::start_node_attempt` wraps it with a parent version bump. Invalid transitions return `InvalidTransition` instead of being swallowed by `let _ = transition_to(Ready)`. On error the node is routed through `mark_setup_failed` rather than dispatched on stale `Failed` state.
- **#301 — panicked tasks report their real NodeKey.** `run_frontier` tracks a `HashMap<tokio::task::Id, NodeKey>` alongside the `JoinSet`, uses `join_next_with_id()`, and looks up the real `NodeKey` via `JoinError::id()`. Panicked nodes are marked `Failed`, checkpointed, and reported via `NodeFailed` with the correct identity instead of a sentinel.
- **#311 — workflow input survives resume.** `ExecutionState` gains `Option<Value> workflow_input` (with `#[serde(default)]` for legacy states). `execute_workflow` and `replay_execution` set it at execution start; `resume_execution` reads it back so entry nodes that crashed mid-run receive the original trigger payload instead of `Null`. Legacy states fall through to `Null` with a WARN log.
- **#299 — idempotency replay preserves routing.** `check_and_apply_idempotency` no longer synthesizes `ActionResult::Success` on replay. It loads the full persisted `ActionResult` via new `ExecutionRepo::save_node_result` / `load_node_result` hooks (default no-op so existing backends compile unchanged; `InMemoryExecutionRepo` overrides both), and hands it to `process_outgoing_edges`. `Branch`/`Route`/`MultiOutput` semantics survive idempotency replay; previously every branch edge fired unconditionally.

## Adaptations against current main

The original #346 diff predated several refactors. Changes this PR absorbed (beyond the direct conflicts):

1. **`NodeId` → `NodeKey` throughout.** PR #412 renamed the identifier; the in-memory `HashMap<tokio::task::Id, NodeId>` became `HashMap<tokio::task::Id, NodeKey>`, `NodeId::new()` sentinels became `node_key!("...")` in tests, and every consumed `NodeKey` needed `.clone()` at first use (it's not `Copy` like the old ULID).
2. **`.project/context/*.md` files deleted on main** (`9fe50c7e` / `#398`). Dropped the incoming updates.
3. **HEAD's `run_frontier` wraps `join_next` in a `select!` against `wall_clock_remaining`.** Kept that shape, switched all three drain sites (timeout, cancel, normal) to `join_next_with_id()`, and added `task_nodes.clear()` on drain.
4. **HEAD's Phase 3 match kept a separate `ActionResult::Retry` arm** (#346 folded it into success). Kept the 4-arm shape, added `task_nodes.remove(&task_id)` at each `Ok(…)` entry.
5. **`replay_execution` path updated** — added the `input_overrides.get(&plan.replay_from)` + `set_workflow_input(...)` computation at the top of the pinned-node loop so #311 fires on replay too.

## Test plan (fresh this turn)

- [x] `cargo +nightly fmt --all` — clean (no diff)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo nextest run --workspace` — **3197 passed**, 13 skipped, 0 failed
- [x] All 5 regression tests from the original #346 pass under the NodeKey rename:
  - `setup_failure_checkpoints_execution_state` (#321)
  - `start_node_attempt_rejects_terminal_state` (#300)
  - `panicked_task_reports_real_node_id` (#301)
  - `resume_restores_original_workflow_input` (#311)
  - `idempotency_replay_preserves_branch_routing` (#299)

## Closes

Closes #299
Closes #300
Closes #301
Closes #311
Closes #321

## Context / evidence trail

- Audit doc: [docs/superpowers/specs/2026-04-18-stale-issue-audit-and-real-bugs.md](https://github.com/vanyastaff/nebula/blob/main/docs/superpowers/specs/2026-04-18-stale-issue-audit-and-real-bugs.md) §2 Group C
- Original PR (closed): [#346](https://github.com/vanyastaff/nebula/pull/346)
- Phantom-fix evidence at `main @ 2b205abf`: `crates/engine/src/engine.rs:1546` — `let fake_result = ActionResult::success(output_value);` — [pinned permalink](https://github.com/vanyastaff/nebula/blob/2b205abf/crates/engine/src/engine.rs#L1546)

### DIFF
--- a/crates/engine/src/engine.rs
+++ b/crates/engine/src/engine.rs
@@ -1195,11 +1195,16 @@ impl WorkflowEngine {
                 )
                 .await;
 
-                self.emit_event(ExecutionEvent::NodeFailed {
-                    execution_id,
-                    node_key: node_key.clone(),
-                    error: setup_err_msg.clone(),
-                });
+                if exec_state
+                    .node_state(node_key.clone())
+                    .is_some_and(|ns| ns.state == NodeState::Failed)
+                {
+                    self.emit_event(ExecutionEvent::NodeFailed {
+                        execution_id,
+                        node_key: node_key.clone(),
+                        error: setup_err_msg.clone(),
+                    });
+                }
 
                 if let Some(err_msg) = abort {
                     cancel_token.cancel();
--- a/crates/engine/src/engine.rs
+++ b/crates/engine/src/engine.rs
@@ -434,6 +434,15 @@ impl WorkflowEngine {
         // the version move (issue #255).
         let mut exec_state = ExecutionState::new(execution_id, workflow.id, &node_ids);
         exec_state.transition_status(ExecutionStatus::Running)?;
+        // Use override inputs if provided — computed up front so the
+        // same value is persisted on the execution state (issue #311)
+        // and fed to the frontier loop.
+        let input = plan
+            .input_overrides
+            .get(&plan.replay_from)
+            .cloned()
+            .unwrap_or(serde_json::Value::Null);
+        exec_state.set_workflow_input(input.clone());
         for node_key in &pinned {
             // NOTE: errors are intentionally discarded here. Pinned nodes are
             // forced through Ready→Running→Completed for bookkeeping; the
@@ -462,13 +471,6 @@ impl WorkflowEngine {
             })
             .cloned()
             .collect();
-
-        // Use override inputs if provided.
-        let input = plan
-            .input_overrides
-            .get(&plan.replay_from)
-            .cloned()
-            .unwrap_or(serde_json::Value::Null);
 
         let error_strategy = workflow.config.error_strategy;
 
@@ -566,6 +568,10 @@ impl WorkflowEngine {
         let node_ids: Vec<NodeKey> = workflow.nodes.iter().map(|n| n.id.clone()).collect();
         let mut exec_state = ExecutionState::new(execution_id, workflow.id, &node_ids);
         exec_state.transition_status(ExecutionStatus::Running)?;
+        // Persist the original trigger payload so that resume can
+        // feed entry nodes the same input instead of substituting
+        // Null (issue #311).
+        exec_state.set_workflow_input(input.clone());
 
         // 4b. Persist initial execution state
         let mut repo_version: u64 = 0;
@@ -873,13 +879,22 @@ impl WorkflowEngine {
             .inc();
 
         let error_strategy = workflow.config.error_strategy;
-        // TODO: the original workflow input is not persisted in ExecutionState.
-        // Resume passes Null, which means re-running entry nodes will not receive
-        // the original trigger data. Entry nodes that have already completed are
-        // skipped via idempotency, so this only affects entry nodes that crashed
-        // mid-execution without completing. To fix, persist `workflow_input` in
-        // ExecutionState alongside the execution ID.
-        let workflow_input = serde_json::Value::Null;
+        // Restore the original trigger payload from the persisted
+        // execution state. Legacy states that predate #311 deserialize
+        // the field as `None` — fall back to `Null` with a warning so
+        // the regression is visible in logs.
+        let workflow_input = match exec_state.workflow_input.clone() {
+            Some(v) => v,
+            None => {
+                tracing::warn!(
+                    %execution_id,
+                    "resume: persisted execution state is missing workflow_input; \
+                     falling back to Null — entry nodes that did not complete \
+                     on the original run will receive Null input"
+                );
+                serde_json::Value::Null
+            },
+        };
         let failed_node = self
             .run_frontier(
                 &graph,
@@ -1014,11 +1029,15 @@ impl WorkflowEngine {
             ready_queue.push_back(node_key);
         }
 
-        // In-flight tasks
+        // In-flight tasks + a side map from tokio task id → NodeKey so
+        // that panics (where the inner future's `(NodeKey, _)` payload
+        // is lost) can still be attributed to the real node instead
+        // of a synthesized placeholder (issue #301).
         let mut join_set: JoinSet<(
             NodeKey,
             Result<ActionResult<serde_json::Value>, EngineError>,
         )> = JoinSet::new();
+        let mut task_nodes: HashMap<tokio::task::Id, NodeKey> = HashMap::new();
 
         // Main frontier loop
         loop {
@@ -1087,6 +1106,7 @@ impl WorkflowEngine {
                     input,
                     &activated_edges,
                     &mut join_set,
+                    &mut task_nodes,
                 );
                 if spawned {
                     let action_key = node_map
@@ -1101,10 +1121,19 @@ impl WorkflowEngine {
                     continue;
                 }
 
-                // Node failed during setup (e.g., param resolution).
+                // Node failed during setup (e.g., param resolution, invalid
+                // state machine position, missing node definition). Read the
+                // error message the setup path recorded on the node state;
+                // fall back to a generic message if absent.
+                let setup_err_msg = exec_state
+                    .node_states
+                    .get(&node_key)
+                    .and_then(|ns| ns.error_message.clone())
+                    .unwrap_or_else(|| "node setup failed".to_owned());
+
                 let abort = handle_node_failure(
                     node_key.clone(),
-                    "parameter resolution failed",
+                    &setup_err_msg,
                     error_strategy,
                     graph,
                     outputs,
@@ -1114,6 +1143,25 @@ impl WorkflowEngine {
                     &mut ready_queue,
                     exec_state,
                 );
+
+                // Checkpoint *after* handle_node_failure so the persisted
+                // node state reflects the final resolved state — symmetrical
+                // to the runtime-failure branch below (issue #321).
+                self.checkpoint_node(
+                    execution_id,
+                    node_key.clone(),
+                    outputs,
+                    exec_state,
+                    repo_version,
+                )
+                .await;
+
+                self.emit_event(ExecutionEvent::NodeFailed {
+                    execution_id,
+                    node_key: node_key.clone(),
+                    error: setup_err_msg.clone(),
+                });
+
                 if let Some(err_msg) = abort {
                     cancel_token.cancel();
                     return Some((node_key, err_msg));
@@ -1127,7 +1175,8 @@ impl WorkflowEngine {
 
             if cancel_token.is_cancelled() {
                 join_set.abort_all();
-                while join_set.join_next().await.is_some() {}
+                while join_set.join_next_with_id().await.is_some() {}
+                task_nodes.clear();
                 break;
             }
 
@@ -1148,14 +1197,15 @@ impl WorkflowEngine {
             tokio::pin!(sleep_fut);
 
             let join_result = tokio::select! {
-                result = join_set.join_next() => match result {
+                result = join_set.join_next_with_id() => match result {
                     Some(r) => r,
                     None => break,
                 },
                 () = &mut sleep_fut => {
                     cancel_token.cancel();
                     join_set.abort_all();
-                    while join_set.join_next().await.is_some() {}
+                    while join_set.join_next_with_id().await.is_some() {}
+                    task_nodes.clear();
                     return Some((
                         node_key!("_timeout"),
                         "execution budget exceeded: max_duration".to_string(),
@@ -1163,14 +1213,16 @@ impl WorkflowEngine {
                 }
                 () = cancel_token.cancelled() => {
                     join_set.abort_all();
-                    while join_set.join_next().await.is_some() {}
+                    while join_set.join_next_with_id().await.is_some() {}
+                    task_nodes.clear();
                     break;
                 }
             };
 
             // Phase 3: Process the completed task
             match join_result {
-                Ok((node_key, Ok(ActionResult::Retry { .. }))) => {
+                Ok((task_id, (node_key, Ok(ActionResult::Retry { .. })))) => {
+                    task_nodes.remove(&task_id);
                     // ActionResult::Retry has no scheduler yet; treat it as a node
                     // failure for at-least-once semantics (#290/#296 short-term).
                     total_retries.fetch_add(1, Ordering::Relaxed);
@@ -1216,7 +1268,8 @@ impl WorkflowEngine {
                         });
                     }
                 },
-                Ok((node_key, Ok(action_result))) => {
+                Ok((task_id, (node_key, Ok(action_result)))) => {
+                    task_nodes.remove(&task_id);
                     mark_node_completed(exec_state, node_key.clone());
 
                     // Track output size for budget enforcement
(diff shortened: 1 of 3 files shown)


vanyastaff:
@copilot apply changes based on the comments in [this thread](https://github.com/vanyastaff/nebula/pull/428#pullrequestreview-4134181452)

Issue #336, opened 2026-04-14
[HIGH] resume_execution unconditionally activates all outgoing edges for completed nodes


vanyastaff:
## Summary
`resume_execution` reconstructs `activated_edges` for terminal `Completed|Skipped` nodes by unconditionally activating **all** outgoing edges, ignoring per-edge conditions and routing metadata.

## Impact
- Control-flow divergence after restart: branches/routes/conditional paths can execute that were never activated before crash.
- Potential duplicate/incorrect side effects in wrong downstream nodes.
- Resume is not behaviorally equivalent to uninterrupted execution.

## Failure Scenario
1. Node completes with selective activation (e.g. `Branch`, `Route`, `OnResult`, `Expression`).
2. Process crashes before workflow fully finishes.
3. During resume, reconstruction marks all outgoing edges from that completed node as activated.
4. Downstream nodes that should stay inactive become ready and execute.

## Evidence
- `crates/engine/src/engine.rs` (`resume_execution`, terminal reconstruction loop):
  - for each terminal node:
    - always marks outgoing as `resolved`,
    - for `Completed|Skipped` unconditionally inserts source into `activated_edges`.
- This bypasses `evaluate_edge(...)` semantics used in normal execution, which depends on result/error and edge condition.

## Why Tests/Lints Missed It
- Resume tests currently focus on basic completed-node replay and not conditional/routed edge scenarios.
- Static checks cannot reason about replaying dynamic edge-activation semantics.

## Proposed Fix
- Persist sufficient edge-activation metadata (or original `ActionResult` shape) for completed nodes.
- On resume, reconstruct activation using persisted decision data, not blanket activation.
- Add regression tests for resume correctness with:
  - `Branch` edges,
  - `Route` / `from_port`,
  - `OnResult` and `Expression` conditions.

## Acceptance Criteria
- Resume produces identical downstream activation set as uninterrupted run for conditional graphs.
- No previously inactive branch executes solely due to restart.

## Related Files
- `crates/engine/src/engine.rs`
- `crates/workflow/src/graph.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

Issue #324, opened 2026-04-14
[HIGH] resume_execution loses historical OnError edge activations from failed nodes


vanyastaff:
## Summary
`resume_execution()` reconstructs `activated_edges` incorrectly for previously failed nodes: it only marks outgoing edges as activated for `Completed|Skipped`, dropping historical `OnError` activations from `Failed` nodes.

## Impact
- Error-handler branches (`OnError`) can be mis-scheduled after restart.
- A handler node may run with empty activation context (and wrong input) or be skipped incorrectly, diverging from pre-crash control flow.
- Breaks crash-consistency guarantees for error-routing workflows.

## Failure Scenario
1. Node `B` fails, `OnError` edge `B -> C` activates, but process crashes before `C` completes.
2. Persisted state contains `B=Failed` and non-terminal `C`.
3. On resume, reconstruction marks edge `B -> C` as resolved but **not activated**.
4. `C` can still become seed (all predecessors resolved), but receives input as if no predecessor activated (workflow-level/null path), not the expected error-driven path.

## Evidence
- `crates/engine/src/engine.rs` (`resume_execution` reconstruction loop):
  - always sets `resolved_edges` for terminal nodes,
  - sets `activated_edges` only for `NodeState::Completed | NodeState::Skipped`.
- `process_outgoing_edges()` semantics:
  - failed nodes can activate `EdgeCondition::OnError`,
  - activation state directly drives input resolution in `resolve_node_input_with_support()`.

## Why Tests/Lints Missed It
- Existing resume tests focus on completed-node replay and generic frontier recovery.
- No restart test for `OnError`-driven branch mid-flight.
- Static checks do not validate replay semantics of dynamic edge activation.

## Proposed Fix
- Persist per-edge activation decisions (or enough replay metadata), or
- deterministically recompute activation for failed terminal nodes using stored failure context and edge conditions.
- Add regression test: crash after failure-handler activation, then resume and verify handler receives error-path input.

## Acceptance Criteria
- Resume preserves `OnError` activation behavior across crashes.
- Handler node executes/skips exactly as in non-crash execution.

## Related Files
- `crates/engine/src/engine.rs`
- `crates/workflow/src/graph.rs`

## Notes on This Audit Pass
- Not fixed in this pass (tracking issue only per audit constraints).

Issue #308, opened 2026-04-14
[HIGH] Runtime execute_stateful: handler state lives only on the stack — crash loses all progress, no checkpoint/resume


vanyastaff:
## Summary

`ActionRuntime::execute_stateful` holds the stateful handler's `state` as a plain stack local and loops through `handler.execute(&input, &mut state, &context)` without persisting between iterations. The code comments acknowledge the gap explicitly:

```rust
// crates/runtime/src/runtime.rs ~264-268
/// Execute a stateful handler — loops through [`StatefulHandler::execute`]
/// with in-memory state checkpointing.
///
/// State persistence is post-MVP — state only lives on the stack of this
/// call.
async fn execute_stateful(
    &self,
    metadata: &ActionMetadata,
    handler: Arc<dyn StatefulHandler>,
    input: serde_json::Value,
    context: ActionContext,
) -> Result<ActionResult<serde_json::Value>, ActionError> {
    // ...
    let mut state = handler.init_state()?;

    for _iteration in 0..MAX_ITERATIONS {
        // ... loop, mutate state, never persist ...
    }
}
```

This is explicitly labelled \"post-MVP\". This tracking issue exists so the gap does not stay hidden behind a code comment.

## What's actually broken

1. **Process crash mid-iteration** — all iteration progress is lost. The next start calls `handler.init_state()` fresh. Any side effects the handler already performed on the external world are not reconciled.

2. **Restart across shutdown** — same as crash but for graceful restart / deploy. A stateful action in iteration 147 of 10,000 resumes at iteration 0 after restart.

3. **Replay semantics undefined** — the replay path (`crates/execution/src/replay.rs`) knows how to pin stateless outputs. Stateful actions have no notion of \"pinned mid-iteration state\" because there's no mid-iteration state to pin.

4. **`JournalEntry::NodeRetrying` never fires for stateful handlers** — retries go through `ActionResult::Continue { delay }`, which loops the handler; there is no attempt boundary to journal.

5. **Idempotency is at node level, not iteration level** — the engine's `check_and_apply_idempotency` treats a stateful node as a single unit. On restart, either the whole stateful node re-runs from scratch or is skipped wholesale; no intermediate granularity.

## Concrete scenarios

[...]

PR #436, opened 2026-04-18
fix(engine): persist before edge routing in run_frontier failure branches (#297)

### LINKED ISSUE #297: [HIGH] engine checkpoint ordering: emit_event + idempotency recording happen around persist — crash between steps causes duplicate execution
## Summary

`run_frontier` emits `ExecutionEvent::NodeCompleted` *before* calling `checkpoint_node`, and calls `record_idempotency` *after* `checkpoint_node`. Every one of those boundaries is a crash window where persist state diverges from \"what already happened\":

```rust
// crates/engine/src/engine.rs ~1064-1098
mark_node_completed(exec_state, node_id);           // 1. in-memory
self.emit_event(ExecutionEvent::NodeCompleted { ... }); // 2. observers see \"done\"
// ... budget bookkeeping ...
self.checkpoint_node(...).await;                    // 3. persist
self.record_idempotency(...).await;                 // 4. persist dedup
process_outgoing_edges(...);                        // 5. frontier advances
```

### A — Crash between `emit_event` and `checkpoint_node`

External observer (TUI, webhook subscriber, metrics sink) receives `NodeCompleted`. They query the state store and see the node still `Running` (no checkpoint yet). Or worse: they trigger a downstream workflow based on the event, and that workflow relies on `outputs` which were not persisted — downstream reads stale data.

### B — Crash between `checkpoint_node` and `record_idempotency`

Checkpoint says \"node Completed\". Idempotency store does NOT have the node's key. On restart, the engine hydrates execution state, sees the node Completed, and... actually the `check_and_apply_idempotency` path is probably skipped because the state says Completed. Let me re-read.

Looking at the seed path: `run_frontier` receives a `seed_nodes: Vec<NodeId>` for resume. The engine rebuilds `ready_queue` from the resumed state. If the seed includes a node whose state is already Completed, `check_and_apply_idempotency` *should* short-circuit. But if the engine's resume logic seeds nodes by consulting `resolved_edges` and a node was one-step-short of being fully resolved when the crash happened, the same node may get re-enqueued.

More importantly: the `IdempotencyManager` (`crates/execution/src/idempotency.rs`) is **in-memory only**. `check_and_apply_idempotency` in the engine consults *some* persistent source (not the in-memory manager — otherwise resume wouldn't dedup at all). But the `record_idempotency` call on line 1085 post-dates the checkpoint, so on restart the persistent dedup table lags the persisted execution state by exactly one node.

### C — Error path has the same shape — and it's worse

On failure (line 1100-1127):

```rust
mark_node_failed(exec_state, node_id, err);
self.emit_event(ExecutionEvent::NodeFailed { ... });
let abort = handle_node_failure(...);  // mutates activated_edges, may route OnError
self.checkpoint_node(...).await;       // persist AFTER routing
if let Some(err_msg) = abort { return; }
```

`handle_node_failure` may activate `OnError` edges and **push successor nodes into `ready_queue`**. If the process crashes between line 1121 and line 1126, the persisted state has `node_id: Running` (from before `mark_node_failed`), but the runtime had already advanced the frontier to a successor. On restart, the engine re-executes `node_id`, and may or may not also re-execute the successor depending on how seed nodes are computed.

## Why this is HIGH

`run_frontier` is the entire execution loop. Every node completion and failure goes through these ordering points. Any of them is a crash-inconsistency window. The symptoms are the classic duplicate-side-effect pattern: payments charged twice, webhooks fired twice, emails sent twice, exactly the kind of bug that's invisible in development and catastrophic in production.

## Fix sketch

Commit the persist step **before** anything external can see the transition:

```rust
// Correct order: persist, THEN announce, THEN move the frontier.
self.checkpoint_node(...).await?;               // 1. durable
self.record_idempotency(...).await?;            // 2. durable dedup
self.emit_event(ExecutionEvent::NodeCompleted { ... }); // 3. observers
process_outgoing_edges(...);                    // 4. frontier
```

For the error path, move `checkpoint_node` *inside* `handle_node_failure` (or have `handle_node_failure` return a \"staged state\" that the caller commits before routing edges), so that the edges the engine activates are a deterministic function of the persisted state.

Additionally, consider making `emit_event` async and make it part of the \"durable\" phase when the event sink is supposed to be transactional (e.g., \"only mark notified after consumer ack\"). Right now events ride a bounded `mpsc` — drops are tolerated, which means event delivery is best-effort even without a crash.

## Severity

**HIGH** — classic duplicate-processing lineage, reachable on every crash/restart.

## Related

- `#261` — `IdempotencyManager::seen` unbounded (separate issue; this one is about ordering, not memory)
- `#283` — resume bypass of transition_to
- `#266` — hardcoded `:1` idem key (masks the ordering issue further)

vanyastaff:
## Summary

Closes D2 of the engine-lifecycle canon cluster ([`docs/plans/engine-lifecycle-canon-cluster-2026-04.md`](docs/plans/engine-lifecycle-canon-cluster-2026-04.md) §Group D, tech-lead sign-off 2026-04-18):

- Split `handle_node_failure` into pure/ordered steps — `classify_failure` → `apply_failure_recovery` → `route_failure_edges` — so edge routing runs as a function of persisted state.
- Reorder all three failure branches in `run_frontier` (setup-failure, retry-as-failure, runtime-failure) to `mark → classify → apply → checkpoint → emit → route`, matching the success path that already runs `mark → checkpoint → record_idempotency → emit → process_edges`.
- Make `checkpoint_node` return `Result<(), EngineError>` with a new `EngineError::CheckpointFailed { node_key, reason }` variant (category: Internal, code: `ENGINE:CHECKPOINT_FAILED`). All four call sites abort the node's progression on `Err` (cancel token + early return) instead of logging-and-continuing.

Closes [#297](https://github.com/vanyastaff/nebula/issues/297).

## Verification finding (per the chip's first-task instruction)

The task brief asked to read `engine.rs:~1064-1098` (from issue #297 body) and `engine.rs:1219-1262` (per explorer) and reconcile. Finding:

| Branch | Pre-this-PR state | Action |
|---|---|---|
| Success path (`Ok((_, Ok(action_result)))`) | **Already correct** — PR [#326](https://github.com/vanyastaff/nebula/pull/326) (commit `a130ea93`, April 14) fixed `mark → checkpoint → record_idempotency → emit → edges`. | Added `?`-style abort on checkpoint Err. |
| Setup-failure (`spawn_node` returned `false`) | Partially fixed by PR [#425](https://github.com/vanyastaff/nebula/pull/425) (D1): `handle_node_failure → checkpoint → emit`. Edge routing still ran before checkpoint. | Reordered to `classify → apply → checkpoint → emit → route`. |
| Retry-as-failure (`Ok((_, Ok(ActionResult::Retry)))`) | `mark_failed → handle_node_failure → checkpoint → emit`. Edge routing ran before checkpoint. | Same reorder. |
| Runtime-failure (`Ok((_, Err))`) | Same shape as retry-as-failure. | Same reorder. |

Issue #297 closed auto on PR #326 merge (squash-merge ghost close); the failure-branch half of the fix had not shipped. This PR closes the remaining scope.

## Why this was a §11.5 / §12.4 / §13 violation

`handle_node_failure` invoked `process_outgoing_edges` internally: for FailFast/ContinueOnError it activated OnError handlers; for IgnoreErrors it routed a null-success payload. Either way, successors entered `ready_queue` — an in-memory structure — before `checkpoint_node` committed the new `NodeState`. A crash between the two lost both:

1. The persisted Failed (or IgnoreErrors-recovered Completed) state → resume re-dispatches the node.
2. Every OnError / ContinueOnError edge-routing decision that was already in memory.

On resume the engine rebuilds `ready_queue` from persisted state; a Running node gets re-executed, duplicating any non-idempotent side effect. Per PRODUCT_CANON §13 knife #5, that is the bar this PR exists to meet.

`checkpoint_node` additionally swallowed three distinct storage failure modes (`save_node_output` error, `transition()` error, CAS mismatch) via `tracing::warn!`. Observers saw `NodeCompleted` / `NodeFailed` for transitions the store never committed — the exact §12.4 silent-failure pattern the task forbids. Surfacing the error as `EngineError::CheckpointFailed` and aborting node progression closes this hole. CAS-mismatch now returns `Err` too (was previously "re-read version and continue"); silently advancing on foreign mutation would let two engine workers diverge on the same execution per §11.1.

## Tests

Three crash-window regression tests + a reusable `FailAtTransitionN` wrapper repo (fails the Nth `transition()`, forwards every other trait method including the stateful-checkpoint trio):

- `runtime_failure_checkpoint_error_aborts_before_edge_routing` — workflow A (fails) --OnError--> B. Inject storage failure on the first `transition()` (A's checkpoint). Asserts **no `NodeStarted(B)` event fires** and **no `NodeFailed(A)` event fires** — observers never learn of A's transition because the store refused it.
- `setup_failure_checkpoint_error_aborts_before_edge_routing` — same contract, parameter-resolution path.
- `ignore_errors_persists_recovered_completed_state` — workflow with one Failed-then-IgnoreErrors-recovered node. Asserts persisted state is `completed` at `>= v3` (create + recovery checkpoint + final). Pre-fix the recovery only persisted at the final-state flush, so a crash between routing-as-success and final flush could lose the override.

All three verified failing on the pre-diff ordering, passing after.

## Canon anchors

- PRODUCT_CANON §11.5 — durability precedes visibility.
- PRODUCT_CANON §12.4 — no silent log-and-continue on state transitions.
- PRODUCT_CANON §13 knife #5 — non-idempotent side effects under restart.
- Architectural-fit verdict: 🟢 — ordering correction within an existing function; no new L2 invariant; no new public surface.
- No ADR required (tech-lead sign-off Q3).

## Behavior change note for §17 DoD reviewer

Behavior-changing PR even though no schema moves. Pre-fix, a storage-layer failure during `checkpoint_node` was a silent WARN and the workflow appeared to complete. Post-fix, it aborts the node's progression and the execution ends in `Failed` with an `ENGINE:CHECKPOINT_FAILED` code. Operators monitoring storage-error dashboards will see surfaced errors where they previously saw log lines; execution counts of "succeeded despite checkpoint failure" become zero.

## Cluster context

- **D1 ([#321](https://github.com/vanyastaff/nebula/issues/321))** — setup-failure checkpoint symmetry: already landed via PR [#425](https://github.com/vanyastaff/nebula/pull/425). This PR inherits the branch D1 established and tightens its ordering.
- **D3 ([#308](https://github.com/vanyastaff/nebula/issues/308))** — stateful contract: separate PR with ADR. Unaffected here.
- **Group B** (resume correctness — [#311](https://github.com/vanyastaff/nebula/issues/311), [#324](https://github.com/vanyastaff/nebula/issues/324), [#336](https://github.com/vanyastaff/nebula/issues/336), [#299](https://github.com/vanyastaff/nebula/issues/299)) is **unblocked** by this PR — per the plan, B's resume-extension work inherits D2's persist-then-announce ordering.

## Review notes

Ran `pr-review-toolkit:silent-failure-hunter` on the diff. One pre-existing silent path not in D2 scope was flagged and acknowledged: `record_idempotency` still `tracing::warn!`'s on `mark_idempotent` failure. That is the success-path dedup contract; §12.4 honesty there is a separate follow-up, not part of D2's checkpoint ordering. Flag for reviewers if you want it addressed in this PR anyway.

## Test plan

- [x] `cargo +nightly fmt --all -- --check` — clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — 0 warnings.
- [x] `cargo nextest run --workspace` — 3195 tests run: 3195 passed, 13 skipped.
- [x] `cargo test --workspace --doc` — passing.
- [x] `lefthook` pre-push (shear, check-all-features, check-no-default, docs, doctests, nextest 3202 passed) — all green.
- [x] Three new regression tests verified failing on pre-diff ordering before implementation, passing after.

### DIFF
--- a/crates/engine/src/engine.rs
+++ b/crates/engine/src/engine.rs
@@ -1150,9 +1150,45 @@ impl WorkflowEngine {
                 // Node failed during setup (e.g., param resolution).
                 // `spawn_node` already marked the node as Failed and stored
                 // the typed error message on `NodeExecutionState`.
-                let abort = handle_node_failure(
+                //
+                // Ordering (§11.5, #297): classify outcome → apply
+                // IgnoreErrors recovery → persist → emit observable
+                // event → route edges. Route AFTER checkpoint so
+                // OnError successors are a function of persisted state.
+                let err_msg = exec_state
+                    .node_state(node_key.clone())
+                    .and_then(|ns| ns.error_message.clone())
+                    .unwrap_or_else(|| "parameter resolution failed".to_string());
+
+                let outcome = classify_failure(error_strategy);
+                apply_failure_recovery(outcome, node_key.clone(), exec_state, outputs);
+
+                if let Err(e) = self
+                    .checkpoint_node(
+                        execution_id,
+                        node_key.clone(),
+                        outputs,
+                        exec_state,
+                        repo_version,
+                    )
+                    .await
+                {
+                    cancel_token.cancel();
+                    return Some((node_key, e.to_string()));
+                }
+
+                if outcome == FailureOutcome::Fail {
+                    self.emit_event(ExecutionEvent::NodeFailed {
+                        execution_id,
+                        node_key: node_key.clone(),
+                        error: err_msg.clone(),
+                    });
+                }
+
+                let abort = route_failure_edges(
+                    outcome,
                     node_key.clone(),
-                    "parameter resolution failed",
+                    &err_msg,
                     error_strategy,
                     graph,
                     outputs,
@@ -1165,37 +1201,6 @@ impl WorkflowEngine {
                 if let Some(err_msg) = abort {
                     cancel_token.cancel();
                     return Some((node_key, err_msg));
-                }
-
-                // Mirror the runtime-failure branch (§11.5, #297/#321):
-                // when the node remains Failed after handle_node_failure
-                // (i.e., not recovered by IgnoreErrors), persist the
-                // failure decision and any OnError/ContinueOnError
-                // edge-routing it triggered before any observer sees the
-                // node as done. Without this, a crash between here and
-                // the final-state checkpoint would lose both the Failed
-                // state and the edge-routing already applied in memory.
-                if exec_state
-                    .node_state(node_key.clone())
-                    .is_some_and(|ns| ns.state == NodeState::Failed)
-                {
-                    self.checkpoint_node(
-                        execution_id,
-                        node_key.clone(),
-                        outputs,
-                        exec_state,
-                        repo_version,
-                    )
-                    .await;
-                    let err = exec_state
-                        .node_state(node_key.clone())
-                        .and_then(|ns| ns.error_message.clone())
-                        .unwrap_or_else(|| "parameter resolution failed".to_string());
-                    self.emit_event(ExecutionEvent::NodeFailed {
-                        execution_id,
-                        node_key: node_key.clone(),
-                        error: err,
-                    });
                 }
             }
 
@@ -1269,7 +1274,37 @@ impl WorkflowEngine {
                         ),
                     ));
                     mark_node_failed(exec_state, node_key.clone(), &err);
-                    let abort = handle_node_failure(
+
+                    // Ordering (§11.5, #297): classify → apply recovery →
+                    // persist → emit → route. Identical shape to the
+                    // runtime-failure branch below; see its comment block.
+                    let outcome = classify_failure(error_strategy);
+                    apply_failure_recovery(outcome, node_key.clone(), exec_state, outputs);
+
+                    if let Err(e) = self
+                        .checkpoint_node(
+                            execution_id,
+                            node_key.clone(),
+                            outputs,
+                            exec_state,
+                            repo_version,
+                        )
+                        .await
+                    {
+                        cancel_token.cancel();
+                        return Some((node_key.clone(), e.to_string()));
+                    }
+
+                    if outcome == FailureOutcome::Fail {
+                        self.emit_event(ExecutionEvent::NodeFailed {
+                            execution_id,
+                            node_key: node_key.clone(),
+                            error: err.to_string(),
+                        });
+                    }
+
+                    let abort = route_failure_edges(
+                        outcome,
                         node_key.clone(),
                         &err.to_string(),
                         error_strategy,
@@ -1284,25 +1319,6 @@ impl WorkflowEngine {
                     if let Some(err_msg) = abort {
                         cancel_token.cancel();
                         return Some((node_key.clone(), err_msg));
-                    }
-
-                    if exec_state
-                        .node_state(node_key.clone())
-                        .is_some_and(|ns| ns.state == NodeState::Failed)
-                    {
-                        self.checkpoint_node(
-                            execution_id,
-                            node_key.clone(),
-                            outputs,
-                            exec_state,
-                            repo_version,
-                        )
-                        .await;
-                        self.emit_event(ExecutionEvent::NodeFailed {
-                            execution_id,
-                            node_key: node_key.clone(),
-                            error: err.to_string(),
-                        });
                     }
                 },
                 Ok((node_key, Ok(action_result))) => {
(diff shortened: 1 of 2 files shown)


vanyastaff:
Both Copilot review findings addressed in [57fb7b4c](https://github.com/vanyastaff/nebula/commit/57fb7b4c).

**1. OnError input payload persistence.** Reordered the failure branches from `mark → apply → checkpoint → emit → route` to `mark → apply → route → checkpoint → emit`. The new order means `route_failure_edges` writes `outputs[failed_node] = {error, node_id}` BEFORE checkpoint, so `resume_execution`'s `load_all_outputs` finds the payload on replay. Successors enqueued into `ready_queue` by `process_outgoing_edges` are in-memory only and do not dispatch until Phase 1 of the next loop iteration — that runs strictly after the checkpoint, so §11.5 "nothing externally observable precedes persist" still holds. If checkpoint returns `Err`, `cancel_token.cancel()` + early return discards the `ready_queue` mutations; no successor spawns.

**2. `override_node_state` Result surfacing.** `apply_failure_recovery` now returns `Result<(), EngineError>` and uses `?` (auto-converting `ExecutionError → EngineError::Execution`). All three failure branches in `run_frontier` abort node progression on `Err` (same pattern as checkpoint errors). No more `let _ = ...` silencing of a real consistency failure.

**Regression coverage.** Added `on_error_payload_is_persisted_before_checkpoint_commits`: runs `A (fails) --OnError--> B`, then reads `repo.load_node_output(A)` directly and asserts the persisted payload carries both `error` and `node_id` fields. Test would fail under the pre-fix ordering.

**Existing invariants preserved.** `failing_node_stops_execution` (failing node with no OnError edges) still asserts `outputs[failed_node] = None` — the payload is only written when `error_handled = true` from `process_outgoing_edges`, same as pre-diff semantics.

### Gate

- `cargo +nightly fmt --all -- --check` — clean.
- `cargo clippy --workspace --all-targets -- -D warnings` — 0 warnings.
- `cargo nextest run --workspace` — **3203 passed**.
- `cargo nextest run -p nebula-engine` — **115 passed**.
- `lefthook` pre-push mirror — **3214 passed** (includes shear, check-all-features, check-no-default, docs, doctests).

Issue #284, opened 2026-04-14
[MEDIUM] Sandbox capability path check falls back to lexical normalization — symlink-to-nonexistent bypass


vanyastaff:
## Summary

`crates/sandbox/src/capabilities.rs:194` validates that a requested path is within an allowed root by trying `canonicalize()` first and falling back to a lexical `normalize_lex` when either side fails to canonicalize. Lexical normalization does not resolve symlinks.

## Attack

1. Attacker has write access to a sandboxed path, e.g. `/allowed/tmp/`.
2. Attacker creates a symlink `/allowed/tmp/escape -> /etc` that points to a directory which does not yet exist (or is not yet accessible to the canonicalize call).
3. Attacker asks the plugin to access `/allowed/tmp/escape/passwd`.
4. `canonicalize(/allowed/tmp/escape/passwd)` fails (target nonexistent).
5. The fallback `normalize_lex` returns `/allowed/tmp/escape/passwd` unchanged (no symlink resolution).
6. The prefix check against `/allowed/tmp/` passes.
7. When the plugin actually `open()`s the path, the kernel resolves the symlink and serves `/etc/passwd`.

This is a standard TOCTOU-via-lexical-path bypass.

## Location

`crates/sandbox/src/capabilities.rs:194` (fallback branch in the canonicalize path check)

## Fix sketch

1. Canonicalize the deepest existing ancestor of the requested path.
2. Reject the request if any ancestor cannot be canonicalized (EACCES / ENOTDIR / ELOOP).
3. Check the canonicalized ancestor is within the allowed root — then append the remaining non-existent suffix.
4. Additionally forbid `..` segments in the non-existent suffix so an attacker cannot craft `canonical/../../etc/passwd`.

Alternatively: use `openat(O_NOFOLLOW)` once and pass the resulting file descriptor into the plugin via an FD-passing channel, avoiding path-based TOCTOU entirely.

## Severity

MEDIUM on the current in-process sandbox (the sandbox itself is a convention — see pitfalls.md), but **HIGH** once Phase 3 OS sandboxing lands because this code then becomes the actual trust boundary. Worth fixing before the boundary hardens.

## Related

Finding `EX-C4-03` from `.project/audit/2026-04-14-deep-review.md`.

Issue #260, opened 2026-04-14
[MEDIUM] Plugin can hijack sibling plugin sockets via forged handshake line


vanyastaff:
**Finding:** EX-C1-05 (deep review 2026-04-14)
**Crate:** nebula-plugin-sdk / nebula-sandbox
**File:** `crates/sandbox/src/process.rs:213` + `crates/plugin-sdk/src/transport.rs:151` (dial)
**Severity:** MEDIUM · **Confidence:** HIGH

## Problem
```rust
// host
transport::dial(handshake_line.trim()).await
// dial:
let line = handshake_line.trim();
let mut parts = line.splitn(3, '|');
// ...
match kind {
    "unix" => { let stream = tokio::net::UnixStream::connect(addr).await?; ... }
```

`dial` accepts whatever path the **child process** prints to stdout as the handshake. A compromised plugin binary could print `NEBULA-PROTO-2|unix|/run/some-other-plugin/sock` and the host happily connects there, talking the duplex protocol to a sibling plugin (or any management socket reachable from the host).

The handshake line is trusted blindly. 0700/0600 directory perms only protect against other UIDs, not other plugins running as the same user.

## Impact
Cross-plugin pivot: a malicious binary in `plugins/` can hijack any plugin socket the host has reachable. Breaks the isolation story.

## Fix
Host-generated per-plugin socket directory:
1. Host creates a dedicated directory (e.g. via `tempfile::TempDir`), passes its path to the child via an env var.
2. Child must bind its socket inside that directory.
3. Host refuses any handshake address that is not inside the expected directory.

Alternative: host binds the listener itself and passes the FD into the child via `CommandExt::fd_mappings` — child never gets to pick the address at all.

See `.project/audit/2026-04-14-deep-review.md` for full context.

PR #452, opened 2026-04-19
fix(sandbox): path_under symlink bypass + forged handshake (#284, #260)

vanyastaff:
## Summary

Security hardening for `nebula-sandbox`. Two independent findings, bundled because both touch the trust boundary code reviewed together (security-lead scope).

- **#284** — `path_under` lexical fallback let an attacker with write access to a granted directory drop a symlink and bypass the filesystem capability check. Now strict: base must canonicalize, path's deepest existing ancestor must canonicalize under it, `..` banned in either argument, `EACCES`/`ELOOP`/`ENOTDIR` fails closed.
- **#260** — `transport::dial` trusted whatever address the plugin printed in its handshake. Now the host allocates the socket path (tempdir with `0700` on Unix, nonce-based pipe name on Windows), passes it via `NEBULA_PLUGIN_SOCKET_ADDR`/`NEBULA_PLUGIN_SOCKET_KIND`, and verifies the announced `kind|addr` matches exactly before dialling.

Both add typed regression tests (`SandboxError::HandshakeAddrMismatch` variant is new; `#[non_exhaustive]` → no semver break).

## Test plan

- [x] `cargo nextest run -p nebula-sandbox -p nebula-plugin-sdk` — 63 tests, all green (including 6 `cfg(unix)` symlink regression tests and the `broker_smoke` integration tests that exercise the new env-var plugin binding path).
- [x] `cargo clippy -p nebula-sandbox -p nebula-plugin-sdk --all-targets -- -D warnings` clean.
- [x] Lefthook pre-commit (typos, taplo, fmt-check, cargo-deny, clippy) green.
- [ ] Security-lead review of trust-boundary rules (`credential-security-review` skill).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4167,6 +4167,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+ "uuid",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4163,6 +4163,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/crates/plugin-sdk/src/transport.rs
+++ b/crates/plugin-sdk/src/transport.rs
@@ -42,17 +42,125 @@ pub const HANDSHAKE_VERSION: u32 = DUPLEX_PROTOCOL_VERSION;
 #[cfg(unix)]
 const MAX_UNIX_SOCKET_PATH_BYTES: usize = 100;
 
+/// Environment variable the host uses to tell the plugin exactly which
+/// socket address to bind (#260). When set, the plugin MUST bind this
+/// address and announce it in the handshake — no plugin-chosen path.
+///
+/// The corresponding [`ENV_SOCKET_KIND`] indicates `"unix"` or `"pipe"`.
+pub const ENV_SOCKET_ADDR: &str = "NEBULA_PLUGIN_SOCKET_ADDR";
+
+/// Environment variable that pairs with [`ENV_SOCKET_ADDR`] and names the
+/// transport kind (`"unix"` for Unix domain sockets, `"pipe"` for Windows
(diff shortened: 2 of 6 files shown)


vanyastaff:
Follow-up addressing the 3 review comments + restoring green CI (commit 2e90776c).

**Root cause of Linux CI failure:** the new regression test `path_under_rejects_symlink_to_nonexistent_target_outside_base` was exercising a `canonicalize NotFound` path that the walk-up loop was treating as "keep climbing" — but the broken symlink at `base/escape` (pointing at `outside/not-yet`) was indistinguishable from a truly-absent component, so the climb reached `base` itself and matched. Fix: cross-check `canonicalize() -> NotFound` against `symlink_metadata(a)`. If the component exists on disk (i.e. IS a broken symlink), fail closed instead of walking to the parent. Added two regression tests (broken-as-leaf and broken-mid-path).

**Review #1** (`process.rs:644`) — `allocate_host_socket_addr` now prefers `/tmp` over the platform temp dir and skips any candidate whose socket path would exceed `transport::MAX_UNIX_SOCKET_PATH_BYTES` (exposed `pub` for this reuse). Addresses macOS `/var/folders/...` overflowing `sun_path`.

**Review #2** (`transport.rs:78`) — `bind_listener` now fails fast with `io::ErrorKind::InvalidInput` on partial env configuration (one of `NEBULA_PLUGIN_SOCKET_ADDR` / `NEBULA_PLUGIN_SOCKET_KIND` set but not both). Silent fallback to self-allocation would have defeated the #260 handshake-forgery mitigation. Classifier extracted to a pure `classify_env` function so tests don't need to mutate process-global env state (the crate `#![forbid(unsafe_code)]`, so Rust 2024's unsafe env setters are out).

**Review #3** (`capabilities.rs:268`) — see root cause above. Closes the re-introduced #284 bypass.

**Local gate:** `cargo +nightly fmt` + `cargo clippy --workspace -- -D warnings` + `cargo nextest run --workspace` (3278 passed, 13 skipped) all green. Lefthook pre-push ran the full CI mirror (shear, doctests, check-all-features, docs, check-no-default, nextest) — green.

Issue #285, opened 2026-04-14
[MEDIUM] Sandbox ONE_SHOT_ID hardcoded to 1 — stale response from timed-out call matches next request


vanyastaff:
## Summary

`crates/sandbox/src/process.rs:159` declares `const ONE_SHOT_ID: u64 = 1`. Every envelope sent to a plugin uses correlation id `1`. If a dispatch times out (e.g. plugin slow) and the host returns an error, but the plugin later finishes and writes a response to the socket, that response carries `id = 1` — the same id the host will use for its *next* request to the same plugin handle.

## Failure path

1. Host sends envelope `{id: 1, ...work A}` to plugin.
2. Plugin takes longer than `self.timeout`.
3. Host's `timeout(rx.recv())` fires, returns `Err(Timeout)` to caller.
4. Host considers the dispatch failed, plugin handle remains alive (sandbox does NOT currently invalidate the handle on timeout).
5. Plugin finishes work A, writes `{id: 1, result: A}` to the socket.
6. Host immediately reuses the plugin for work B: sends `{id: 1, ...work B}`.
7. Host `recv()` pulls the buffered response for work A, tags it as work B's result.
8. Cross-talk. Work B's caller gets work A's output.

This is correlation-id confusion between distinct in-flight calls, which bidirectional protocol design normally prevents via monotonically increasing request IDs.

## Fix sketch

- Replace `const ONE_SHOT_ID: u64 = 1` with a `Mutex<u64>` or `AtomicU64` counter on `PluginHandle`, incremented per dispatch.
- The receive side matches on the returned `id` and discards any mismatch.
- On timeout, invalidate the plugin handle (close the socket) so no further envelopes can be dispatched through it.

The audit note for `#257` (dispatch_envelope ignores CancellationToken) and this one overlap and should be fixed together — both need a correlation-id overhaul.

## Severity

MEDIUM — low likelihood in practice (timeouts are rare and the handle has a short life), but the failure mode is silent data corruption between workflows, which is the worst category of bug in multi-tenant systems.

## Related

Finding `EX-C3-02` from `.project/audit/2026-04-14-deep-review.md`. Overlaps with `#257`.

Issue #259, opened 2026-04-14
[MEDIUM] Sandbox dispatch retries non-idempotent actions on any failure


vanyastaff:
**Finding:** EX-C2-06 (deep review 2026-04-14)
**Crate:** nebula-sandbox
**File:** `crates/sandbox/src/process.rs:213`
**Severity:** MEDIUM · **Confidence:** HIGH

## Problem
```rust
async fn dispatch_envelope(&self, envelope: HostToPlugin) -> ... {
    let first_attempt = self.try_dispatch(envelope.clone()).await;
    if first_attempt.is_ok() {
        return first_attempt;
    }
    *self.handle.lock().await = None;
    self.try_dispatch(envelope).await
}
```

Retries **every** failure, including non-idempotent action invocations. If the plugin executed a side effect (sent a message, charged a card) and the host failed only on the response read, the host then respawns a fresh plugin and re-runs the same `ActionInvoke` — silently double-invoking the action.

No idempotency token in `HostToPlugin::ActionInvoke`.

## Impact
Silent at-least-twice semantics for any side-effecting plugin action whenever the read leg fails or times out.

## Fix
1. Differentiate "broken pipe before send" vs "broken pipe after send" at the `try_dispatch` call site, only retry on clearly-pre-write errors.
2. OR add an idempotency token to the envelope and let the plugin dedupe.

See `.project/audit/2026-04-14-deep-review.md` for full context.

Issue #257, opened 2026-04-14
[HIGH] ProcessSandbox dispatch_envelope ignores CancellationToken


vanyastaff:
**Finding:** EX-C2-03 (deep review 2026-04-14)
**Crate:** nebula-sandbox
**File:** `crates/sandbox/src/process.rs:223` (try_dispatch)
**Severity:** HIGH · **Confidence:** HIGH

## Problem
```rust
let result = tokio::time::timeout(self.timeout, async {
    handle.send_envelope(&envelope).await?;
    handle.recv_envelope().await
}).await;
```

`dispatch_envelope` does NOT observe the engine's `CancellationToken`. Once the host has sent an envelope, the call is locked to `self.timeout` even if the workflow was cancelled, the action context expired, or the parent was dropped. The `SandboxedContext` is never `select!`'d against.

Worse: `ProcessSandbox::execute` calls `context.check_cancelled()` only at entry and then drops the context — it isn't even passed to `dispatch_envelope`.

## Impact
- Cancelling a long workflow leaves blocked plugin RPCs ticking up to `self.timeout` (seconds). The engine cannot reclaim the slot.
- With a hung plugin and an aggressive timeout this masks itself; with a slow remote it serialises shutdown.
- Violates the `check_cancelled` contract that every other context-aware path in the engine honours.

## Fix
- Thread the `SandboxedContext` (or its `cancellation` token) into `dispatch_envelope`.
- Race the round-trip against `context.cancellation.cancelled()` via `tokio::select!`, surface `ActionError::Cancelled`.
- After cancellation, invalidate the handle — the connection state is undefined mid-cancel (partial write on the wire).

See `.project/audit/2026-04-14-deep-review.md` for full context.

PR #453, opened 2026-04-19
fix(sandbox): honour CancellationToken + monotonic correlation ids (#257, #285)

vanyastaff:
## Summary

Dispatch-core correctness fixes. Both touch `try_dispatch`, bundled so reviewers see the whole dispatch shape in one diff.

- **#257** — `ProcessSandbox::try_dispatch` used plain `tokio::time::timeout`, ignoring the engine's `CancellationToken`. A cancelled workflow had to wait out the full plugin timeout before the slot could be reclaimed. New `race_cancel_timeout` helper races send+recv against both the timeout AND the cancel token (biased select so pre-cancelled tokens are observed on first poll); on cancel, the cached `PluginHandle` is invalidated (mid-write state undefined) and `ActionError::Cancelled` is surfaced.
- **#285** — `ONE_SHOT_ID = 1` violated the protocol's "IDs unique within a plugin process lifetime" contract. Replaced with a per-sandbox `AtomicU64` counter; `try_dispatch` remembers the outbound id and poisons the transport with `SandboxError::ResponseIdMismatch` on any divergence. Necessary groundwork for slice 1d (concurrent multiplexed dispatch).

## Test plan

- [x] `cargo nextest run -p nebula-sandbox` — 41 tests, all green (6 new `race_cancel_timeout_*` + 4 new id-matching tests).
- [x] `cargo clippy -p nebula-sandbox --all-targets -- -D warnings` clean.
- [x] Lefthook pre-commit green.
- [ ] Rust-senior review of `select!` + cancel shape (`rust-senior` agent scope).

## Stacked by

PR #259 (`fix/sandbox-no-blind-retry`) stacks on this branch — rebases to `main` automatically after this merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4181,6 +4181,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-util",
  "tracing",
  "uuid",
 ]
--- a/crates/sandbox/Cargo.toml
+++ b/crates/sandbox/Cargo.toml
@@ -25,6 +25,7 @@ tempfile = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 tokio = { workspace = true, features = ["process", "io-util", "sync", "net"] }
+tokio-util = { workspace = true }
 uuid = { workspace = true }
 
 # OS-level sandbox (Linux only)
--- a/crates/sandbox/src/error.rs
+++ b/crates/sandbox/src/error.rs
@@ -97,4 +97,21 @@ pub enum SandboxError {
         /// The address announced by the plugin in its handshake line.
         got: String,
     },
+
+    /// The plugin's response envelope carried a correlation id that did
+    /// not match the id the host sent in the request (#285). Signals a
+    /// stale response (e.g. a late reply to a cancelled call) or a
+    /// protocol violation. The transport is poisoned after this error
+    /// so the next request respawns the plugin on a fresh connection —
+    /// defense-in-depth against cross-request id confusion.
+    #[error(
+        "plugin response id mismatch: sent {expected}, got {got} — \
+         possible stale response or protocol violation; connection poisoned"
+    )]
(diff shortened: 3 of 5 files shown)


vanyastaff:
Addressed both Copilot review points on `crates/sandbox/src/process.rs` in 444f6414.

**1. Narrowed dispatch retry (line 465).** Introduced a private `TryDispatchError` enum and split the dispatch outcome into `Respawnable` vs `Terminal`. Only `SandboxError::PluginClosed` — stale handle, plugin exited before we sent this envelope — is respawn-eligible; retrying then is safe because no bytes reached a running plugin process. Everything else is terminal on the first attempt and surfaced as-is:

- `ActionError::Cancelled` — never retried; duplicating work the engine asked us to abort would violate the cancellation contract.
- Timeout — classified terminal (was already retryable-flavored, now doesn't re-enter dispatch).
- Protocol violations (`ResponseIdMismatch`, `PluginLineTooLarge`, `HandshakeLineTooLarge`, `TransportPoisoned`, `MalformedEnvelope`, `HostMalformedEnvelope`) — terminal; a retry on a fresh connection would either replay bytes the plugin already processed or give an attacker another shot at the cap.
- Spawn failure — terminal; a binary that just failed to start almost never helps from a quiet retry.

**2. Replaced `expect`.** `guard.as_mut().expect("handle set above")` is now a `let-else` that returns `ActionError::fatal("process sandbox handle missing after successful spawn")`. The path is still unreachable in practice, but a hypothetical logic bug now surfaces through the engine's standard error channel instead of panicking inside the sandbox lock.

Eight new unit tests cover the classification of every `SandboxError` variant plus the `TryDispatchError` round-trip. Full crate gate locally: `cargo +nightly fmt --all`, `cargo clippy -p nebula-sandbox -- -D warnings`, and `cargo nextest run -p nebula-sandbox` (49/49 green, including the 8 new tests). Pre-push hook mirrored CI and all workspace tests (3284) passed before the push.

PR #454, opened 2026-04-19
fix(sandbox): stop blind retry of non-idempotent dispatch (#259)

vanyastaff:
## Summary

`ProcessSandbox::dispatch_envelope` previously ran `try_dispatch` a second time after clearing the cached handle on any error — silently double-invoking non-idempotent `ActionInvoke`s whenever the plugin had already executed the side effect but the response read had failed (#259).

Drop the implicit retry. A stale/dead handle still clears itself inside `try_dispatch` on error, and the next call that reaches `dispatch_envelope` transparently respawns via `spawn_and_dial` — so fault tolerance is preserved, just pushed one level up to the engine's configurable retry policy (the layer that can reason about whether a specific action is safe to retry).

## Test plan

- [x] `cargo nextest run -p nebula-sandbox` — 41 tests, all green.
- [x] `cargo clippy -p nebula-sandbox --all-targets -- -D warnings` clean.
- [x] Lefthook pre-commit green.

## Stacked on

Base is https://github.com/vanyastaff/nebula/pull/453 (#257 + #285) — GitHub will auto-rebase to `main` once that merges. No new tests needed: the change is a removal, verifiable by inspection; existing handle-invalidation tests cover the remaining contract.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4165,6 +4165,7 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-util",
  "tracing",
 ]
 
--- a/crates/sandbox/Cargo.toml
+++ b/crates/sandbox/Cargo.toml
@@ -24,6 +24,7 @@ serde_json = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 tokio = { workspace = true, features = ["process", "io-util", "sync", "net"] }
+tokio-util = { workspace = true }
 
 # OS-level sandbox (Linux only)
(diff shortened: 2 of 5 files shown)


vanyastaff:
Superseded by #453, which merged a strictly stronger fix: the `TryDispatchError::{Respawnable, Terminal}` classifier narrows retry to the provably-safe `SandboxError::PluginClosed` case (stale handle, plugin crashed before any envelope bytes on wire) while keeping fault tolerance. This PR removed all retry entirely, losing the safe respawn-on-stale-handle path. Closing without merge.

PR #460, opened 2026-04-19
fix(credential): close refresh-waiter lost-wakeup race with oneshot per waiter (#268)

### LINKED ISSUE #268: [MEDIUM] Refresh waiter lost-wakeup race not fully closed by Notified::enable()

vanyastaff:
## Summary

Closes [#268](https://github.com/vanyastaff/nebula/issues/268).

The `RefreshAttempt::Waiter` arm relied on `Notify::notify_waiters`, whose contract only wakes waiters *already registered* at the moment the call fires. Between returning `Arc<Notify>` from `try_refresh` (lock released) and the waiter reaching `tokio::pin!(notified); enable();`, a winner that completed a sub-millisecond refresh could race past the registration and the wakeup was lost. Prior audit work narrowed the window to a handful of instructions and capped waiter latency at 5 s, but the race remained.

This PR implements the **Option C** fix recommended in the issue: one `tokio::sync::oneshot` per waiter, registered under the coordinator's map mutex so the race is closed by construction.

### Design

- Winner inserts an `InFlightEntry { senders: Mutex<Vec<Sender>> }` under the coordinator's map mutex and returns `RefreshAttempt::Winner` (now a unit variant — scopeguard calls `complete(id)` directly).
- Waiter, still under the same map mutex, pushes a fresh `oneshot::Sender` into the entry's Vec and returns `RefreshAttempt::Waiter(Receiver)`. Registration is strictly ordered against any future winner's `complete()`.
- `complete()` removes the entry under the map mutex, drains the Vec, and fires every sender. Dropped receivers (cancelled waiters) are ignored via `let _ = tx.send(())`.

The resolver's Waiter arm simplifies to a plain `rx.await` inside the 5 s timeout. The timeout now exists only as belt-and-suspenders against a wedged winner holding the `refresh_semaphore` permit indefinitely, not as a lost-wakeup recovery. The pre-wait `pin!` + `enable()` dance is gone.

### Public API change

- `RefreshAttempt::Winner` is now a unit variant (was `Winner(Arc<Notify>)`).
- `RefreshAttempt::Waiter` wraps `oneshot::Receiver<()>` (was `Waiter(Arc<Notify>)`).

`nebula-credential` is `frontier` API stability per [docs/MATURITY.md](docs/MATURITY.md). Only test callers are affected in the workspace; all updated.

## Tests

- **`waiter_registered_under_lock_is_never_missed`** — registers a waiter and immediately calls `complete()`; the old Notify-based design had a race here, the oneshot-under-lock design does not.
- **`cancelled_waiter_does_not_stall_winner`** — receiver dropped before completion, `complete()` drains cleanly.
- **Obsolete `pre_enabled_notified_receives_notify_waiters`** removed: it pinned the Notify invariant that the new design no longer relies on.

## Test plan

- [x] `cargo nextest run -p nebula-credential` — 325/325 passed
- [x] `cargo clippy -p nebula-credential -- -D warnings` — clean
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] Pre-push hook (mirrors CI): shear, docs, check-all-features, doctests, nextest, check-no-default — all green

## Canon alignment

- PRODUCT_CANON §4.2 Safety (credentials behind abstractions; refresh engine-owned) — strictly preserved. ✓
- PRODUCT_CANON §4.3 Keep-alive (engine durability, process-death tolerance) — waiter latency bound strengthened from "5 s + race" to "5 s hard cap, no race by construction". ✓

### DIFF
--- a/crates/credential/src/refresh.rs
+++ b/crates/credential/src/refresh.rs
@@ -9,10 +9,21 @@
 //! Per-credential refresh state is tracked in a `HashMap` behind a `Mutex`.
 //! The first caller to request a refresh for a given credential ID becomes
 //! the "winner" and performs the refresh. Subsequent callers receive a
-//! [`Notify`](tokio::sync::Notify) handle and wait until the winner completes.
+//! [`tokio::sync::oneshot::Receiver`] and wait until the winner completes.
+//!
+//! # Lost-wakeup safety
+//!
+//! Waiters register a `oneshot::Sender` into the in-flight entry **under
+//! the same mutex** that the winner holds when inserting or removing the
+//! entry. A winner cannot `complete()` past a waiter's registration — the
+//! race window that existed with [`tokio::sync::Notify::notify_waiters`]
+//! (where the waiter had to `enable()` its `Notified` future after the
+//! lock was released) is closed by construction. See GitHub issue #268.
 //!
 //! The winner **must** call [`RefreshCoordinator::complete()`] when done
 //! (success or failure) to wake waiters and remove the in-flight entry.
+//! Wrapping the call in a `scopeguard` is recommended so the guarantee
+//! survives panics and error paths.
 //!
 //! # Examples
 //!
@@ -22,12 +33,12 @@
 //! let coord = RefreshCoordinator::new();
 //!
 //! match coord.try_refresh("cred-1") {
-//!     RefreshAttempt::Winner(_notify) => {
+//!     RefreshAttempt::Winner => {
 //!         // Perform refresh...
 //!         coord.complete("cred-1");
 //!     }
-//!     RefreshAttempt::Waiter(notify) => {
-//!         notify.notified().await;
+//!     RefreshAttempt::Waiter(rx) => {
+//!         let _ = rx.await; // Err means the winner dropped the entry
 //!         // Re-read credential from store
 //!     }
 //! }
@@ -37,7 +48,19 @@ use std::{collections::HashMap, fmt, num::NonZeroUsize, sync::Arc, time::Duratio
 
 use lru::LruCache;
 use nebula_resilience::circuit_breaker::{CircuitBreaker, CircuitBreakerConfig, Outcome};
-use tokio::sync::Notify;
+use tokio::sync::oneshot;
+
+/// In-flight refresh entry.
+///
+/// Holds the pending `oneshot::Sender`s of every waiter registered for this
+/// credential. Waiters push a sender under the coordinator's map lock; the
+/// winner drains and fires every sender on completion.
+struct InFlightEntry {
+    /// Senders to wake on completion. `parking_lot::Mutex` because the list
+    /// is touched only while holding (or having just released) the outer
+    /// map lock, with no `.await` in between.
+    senders: parking_lot::Mutex<Vec<oneshot::Sender<()>>>,
+}
 
 /// Coordinates credential refresh to prevent thundering herd.
 ///
@@ -57,7 +80,7 @@ use tokio::sync::Notify;
 /// // First caller wins
 /// assert!(matches!(
 ///     coord.try_refresh("cred-1"),
-///     RefreshAttempt::Winner(_),
+///     RefreshAttempt::Winner,
 /// ));
 ///
 /// // Second caller waits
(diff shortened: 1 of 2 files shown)


vanyastaff:
Addressed both Copilot review comments on `crates/credential/src/refresh.rs`:

- **Line 41 (module-level example)**: applied the suggested verbatim replacement — `rx.await` returns `Err(oneshot::error::RecvError)` only when the sender is dropped *without sending* (coordinator shutdown, or a bug/cancellation path that prevents `complete()` from running). `Ok(())` is the normal completion signal. Commit: 108358a0.
- **Line 142 (`RefreshAttempt::Waiter` doc)**: already reworded in 3a65997a. The doc now makes clear that `Err(_)` is *not* a normal winner-completion signal; a winner that forgets `complete()` leaves the sender alive in the map, so the receiver hangs until the caller-side timeout rather than erroring.

Verification (run this session):
- `cargo +nightly fmt --all` — clean
- `cargo clippy -p nebula-credential -- -D warnings` — clean
- `cargo test --doc -p nebula-credential` — `25 passed; 0 failed; 21 ignored`
- Full pre-push hook chain (shear, check-all-features, docs, check-no-default, nextest 3269/3269 pass, doctests) — all green locally

No code semantics changed; docs-only.

PR #768, opened 2026-06-01
feat(credential): #[credential] attribute macro — ADR-0088 D1 (one declaration site, capability-by-method)

vanyastaff:
## What

Implements **ADR-0088 D1**: the `#[nebula_credential::credential]` attribute
macro — the single-declaration-site authoring path for a credential — and
dogfoods all three built-ins (api_key, basic_auth, oauth2) onto it.

The author writes ONE inherent `impl` block; the macro reads which methods are
present and emits the base `Credential` impl, one capability sub-trait impl per
capability **method** supplied (`refresh` ⇒ `Refreshable`, `revoke` ⇒
`Revocable`, `test` ⇒ `Testable`, `continue_resolve` ⇒ `Interactive`,
`release` ⇒ `Dynamic`), the five `plugin_capability_report::IsX` consts, and a
`CredentialLifecycle::policy()`.

## Why it's the right shape

Capability is **inferred from method presence, never declared**. There is no
`capabilities(refreshable)` flag that can disagree with the impl — so the
capability-report consts and the lifecycle policy can never lie about what the
type implements. This *strengthens* the `E0046` compile-gate the capability
sub-trait split bought (declaring a capability you didn't implement is now
**unrepresentable**) while collapsing the old four declaration sites
(trait impl + sub-trait impl + `IsX` const + lifecycle policy) into one.

- `policy()` is synthesized from the `category` arg + method presence, and is
  **author-overridable** for state-dependent policies (OAuth2 keeps its
  hand-written `RefreshToken`-while-refresh-token-held / else `ReAcquire`).
- `metadata()` is relocated if hand-written, or synthesized from args.
- A typo'd capability method is **rejected** at the macro site (it cannot
  silently produce a credential missing the capability).

## Scope (expand-contract step 2 of the ADR)

This is additive. The legacy `#[derive(Credential)]` path is untouched and
stays during the migration window; later steps migrate the rest of the
consumers and then delete the old surface.

## Validation

- All 181 credential lib tests + the full oauth2 suite remain green — the
  built-ins' existing tests are the **output oracle** the macro must reproduce
  (including the `Credential + Interactive + Refreshable + Revocable +
  Testable` type-level bound and the refresh-token-presence policy test).
- New positive integration coverage (`credential_attr_macro.rs`) for the paths
  the built-ins don't exercise: capability inference folded through
  `compute_capabilities`, and the synth policy across
  `RefreshToken`/`Lease`/`Static` + `None`/`HandleBased`.
- New compile-fail probes pin the reject diagnostics (typo'd method;
  `continue_resolve` without `type Pending`).
- Downstream consumers (api, engine, plugin, resource) compile; rustdoc
  `-D warnings` green for both crates.

## Commits

1. macro implementation + dogfood static builtins (api_key, basic_auth)
2. dogfood OAuth2 (full capability set + relocated state-dependent policy)
3. macro test coverage (positive + compile-fail)
4. crate-qualify intra-doc links broken by the import move
5. refresh compile-fail `.stderr` for the macro-dogfooded builtins

## Summary by CodeRabbit

* **New Features**
  * Added a `#[credential]` attribute macro for streamlined credential declaration with automatic capability trait inference and metadata synthesis.
  * Refactored credential implementations to use the new macro for cleaner, more maintainable code.

* **Tests**
  * Added comprehensive test coverage for the credential macro, including compile-fail tests and integration tests to validate macro behavior and error handling.

### DIFF
--- a/crates/credential/macros/src/credential_attr.rs
+++ b/crates/credential/macros/src/credential_attr.rs
@@ -203,8 +203,33 @@ fn expand_inner(args: TokenStream2, input: TokenStream) -> syn::Result<TokenStre
         ));
     }
 
+    // A leased/dynamic credential's expiry and renewability live in its state
+    // (lease id + duration), which the macro cannot read. A synthesized
+    // `RefreshStrategy::Lease` policy would therefore carry `lease: None` and
+    // report the credential as non-expiring / non-renewable — a lie for a
+    // Vault/k8s-style secret. Require an explicit `fn policy` in that case.
+    if items.release.is_some() && items.policy.is_none() {
+        return Err(diag::error_spanned(
+            &item.self_ty,
+            "a credential with `fn release` (Dynamic/leased) must hand-write `fn policy` — \
+             the macro cannot infer the lease id/duration from state, so it cannot synthesize \
+             a correct `RefreshStrategy::Lease` policy (set `lease: Some(LeaseRef { … })`)",
+        ));
+    }
+
     let self_ty = &item.self_ty;
     let (impl_generics, _ty_generics, where_clause) = item.generics.split_for_impl();
+
+    // Forward the author's outer attributes (e.g. `#[cfg(...)]`,
+    // `#[allow(...)]`) — minus doc comments — onto every generated impl, so a
+    // cfg-gated or lint-configured annotated block yields cfg-gated /
+    // lint-configured trait impls rather than silently unconditional ones.
+    let outer_attrs: Vec<&Attribute> = item
+        .attrs
+        .iter()
+        .filter(|a| !a.path().is_ident("doc"))
+        .collect();
+    let fwd = quote! { #(#outer_attrs)* };
 
     // ── metadata: relocate the author's, or synthesize from args ──────────
     let metadata_fn = if let Some(m) = &items.metadata {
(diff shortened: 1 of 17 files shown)


vanyastaff:
Addressed all review findings in `50a9865e`:

- **Codex P2 (leased synth policy lies):** `fn release` (Dynamic/leased) without a hand-written `fn policy` is now a **compile error** — the macro can't read the lease id/duration from state, so it refuses to synthesize a `RefreshStrategy::Lease` with `lease: None`. The leased test fixture now hand-writes its policy and asserts correct `is_expiring()`/`is_auto_renewable()`.
- **Codex P2 (`::nebula_core::` unresolved for downstream):** generated metadata now uses the re-exported `::nebula_credential::credential_key!`, so a consumer depending on `nebula-credential` but not `nebula-core` resolves it.
- **CodeRabbit (panic in synth `metadata()`):** the no-icon/no-doc_url case now uses the infallible `CredentialMetadata::for_credential` constructor — no `expect(...)` emitted into library code. The builder+`expect` path remains only for icon/doc_url (the `metadata()` trait method is infallible and the builder is the only icon path; identical to the built-ins and the legacy derive).
- **Copilot (dropped outer attrs):** the annotated block's `#[cfg(...)]`/`#[allow(...)]` (doc comments excluded) are now forwarded onto every generated impl, so cfg-gated/lint-configured blocks produce cfg-gated/configured impls.
- **Copilot (`unreachable!()` in `assoc_type`):** now returns a `syn::Result` — an internal invariant violation surfaces as a normal compile diagnostic, not a proc-macro panic.
- **CodeRabbit (test coverage):** added the inverse compile-fail probe (`type Pending` without `continue_resolve`) and a happy-path Interactive fixture asserting `Capabilities::INTERACTIVE`.

All green: clippy `-D warnings` (`--all-targets`), 181 lib + 5 integration + 3 compile-fail tests, rustdoc `-D warnings`.

Note: the legacy `#[derive(Credential)]` has the same `::nebula_core::credential_key!` pattern, but it's slated for deletion in a later ADR-0088 step and is out of scope here.

PR #769, opened 2026-06-01
refactor(credential)!: collapse 4 credential registries → 1 (ADR-0088 D3)

vanyastaff:
## What

ADR-0088 **D3** registry collapse. Four parallel credential registries become
one capability+metadata source (`CredentialRegistry`) plus the type-erased
async-op table (`DispatchOps<B,PS>`):

| Table | Before | After |
|-------|--------|-------|
| `CredentialRegistry` (credential) | instance + `Capabilities` bitflag + crate | **kept — the one source of truth** |
| `CredentialDispatch` (runtime) | 3 bool flags (refreshable/testable/revocable) | **deleted** — flags mirrored the bitflag |
| `StateProjectionRegistry` (engine) | state-KIND → `project` closure | **deleted** — vestigial (see below) |
| `register_credential_complete` (runtime) | 3-table registrar | **deleted** — test-only, no prod caller |
| `DispatchOps<B,PS>` (runtime) | async op closures | **kept** — must stay generic over store/pending |

Capability is now read solely from the `CredentialRegistry` `Capabilities`
bitflag, computed once at `register::<C>()` from sub-trait membership
(`compute_capabilities`). `CredentialService` capability/`contains` guards
repoint from `self.dispatch.*` to `self.registry.*` — behavior-preserving (the
bitflag derives from the same sub-traits the dispatch flags mirrored).

## Two findings that refined the plan (beyond the original mapping)

1. **`StateProjectionRegistry` was vestigial.** Its type-erased
   `project(state_kind, bytes)` has **zero production callers** — the engine
   resolver is generic over `C` and calls `C::project(&state)` directly
   (resolver.rs:78/127/148/538/686/695). So it was deleted outright rather than
   folded into Core.
2. **State KIND is not unique.** `ApiKeyCredential` and `BearerTokenCredential`
   both use `State = SecretToken` (KIND `secret_token`), so the engine's
   fatal-duplicate-KIND check was unsound. The N7 supply-chain defense
   correctly lives at **KEY** granularity — `CredentialRegistry` dup-KEY stays
   fatal (§15.6) and is unchanged.

## Safety / blast radius

- No production code wires `CredentialServiceBuilder` yet (the facade is
  unbuilt — api-wiring is a later ADR step), so the builder-API change
  (`dispatch` arg removed) has **no external callers**.
- N7 (supply-chain credential takeover via duplicate registration) is
  unchanged — defended at KEY granularity.
- Net **~−690 LoC**; deleted the now-meaningless `registry_sync_invariant` +
  `composite_registration` + `state_projection_duplicate_kind_fatal` tests
  (dup-KEY fail-closed stays covered by `runtime_duplicate_key_fatal`).

## Validation

749 tests pass (credential + runtime + engine, nextest); clippy `-D warnings`
(`--all-targets`) + rustdoc `-D warnings` green; `nebula-api` (downstream)
compiles. Pre-push (clippy-full + crate-diff-gate) green.

## Commits

1. `refactor(credential)!`: collapse `CredentialDispatch` into the capability bitflag
2. `refactor(engine)!`: delete vestigial `StateProjectionRegistry`
3. `docs(credential)`: record the D3 outcome in ADR-0088

## Release Notes

* **Refactor**
  * Consolidated credential capability tracking from multiple registries into a single unified source in `CredentialRegistry`, improving system maintainability and clarity.
  * Simplified internal architecture by removing parallel dispatch and state projection mechanisms.

* **Chores**
  * Removed deprecated composite registration API and associated test infrastructure as part of architecture consolidation.


vanyastaff:
Addressed in `498f7c17`:

**Doc consistency (Copilot ×4 + CodeRabbit ×3)** — fixed all stale `dispatch`/`mark_refreshable` references introduced by the collapse: `TypeCapabilities` rustdoc, the `test_support` module doc, both in-memory fixture-helper docs, and the ADR-0088 D3 lead-in (it no longer says `DispatchOps` "is deleted", which contradicted the retained-async-ops note — now scoped to "its capability role is deleted").

**Codex P2 (`service.rs:972` — discovery may advertise a capability whose ops aren't wired):** valid footgun, but **deferring with reason**:
- Per ADR-0088 D3, capability *is* sub-trait membership — an intrinsic property of the type, computed at registration. Discovery (`list_types`/`get_type`) reporting it is correct; a type that `impl Refreshable` *is* refreshable regardless of whether a given service instance wired the ops. A missing-ops call surfaces as a typed `CapabilityUnsupported` at runtime, not a silent no-op.
- The registry (Core, non-generic) and `DispatchOps<B,PS>` (generic over store/pending) **cannot** be coupled at `CredentialRegistry::register` — the ops need `<B,PS>` that Core never sees. A build-time "registry caps ⊆ ops-registered caps" assertion has to live at the `CredentialServiceBuilder` level, which is reworked in **ADR-0088 D4** (facade merge / non-generic builder). That's the right place to add the consistency gate, so I'm tracking it there rather than bolting a partial check onto the current generic builder.
- Not reachable today: there is no production `CredentialServiceBuilder` caller yet (the facade is unbuilt); the only builders are test fixtures, which register ops correctly.

All green: clippy `-D warnings`, rustdoc `-D warnings`, 749 tests, pre-push.

PR #770, opened 2026-06-01
refactor(credential)!: canonical owner_id derivation — fix tenant-scope format drift (ADR-0088 D7)

vanyastaff:
## What

ADR-0088 **D7** owner_id format-drift fix. The two credential planes derived the
tenant `owner_id` key with **different separators** — the API edge keyed
`"{org}:{workspace}"`, the credential-runtime keyed `"{org}/{workspace}"`. Both
planes now route through one canonical derivation:

- **`Scope::credential_owner_id()`** in `nebula-storage-port` (Core) — the sole
  derivation, callable by every producer (api, runtime, tenancy) on the existing
  `Scope` type, no deny-graph change beyond listing `credential-runtime` on the
  storage-port wrapper (a legal Exec→Core edge).
- `nebula_credential_runtime::TenantScope::new` and `nebula-api`'s
  `owner_id_from_scope` both delegate to it.

## Severity (honest)

**Latent**, not a live incident. The two planes do not share a backend today —
`AppState::with_credential_service` has zero callers and the runtime
`CredentialService` is only constructed in tests. But the moment the facade is
wired (the planned next step), the `:` vs `/` mismatch would silently partition
a single tenant's credentials into two non-intersecting halves — a credential
written via the API invisible to runtime resolution and vice-versa, "silent and
hard to debug". Fixed now as hardening-before-it-bites.

## Encoding (collision-safe)

`org_id`/`workspace_id` are **unvalidated free strings**, so no single separator
is safe: `org="a", ws="b:c"` and `org="a:b", ws="c"` both derive `"a:b:c"` under
`:` (and `/` is identical) — two distinct tenants collapsing to one key. The
owner segment is **length-prefixed** (`{org_len}␞{org}␞{ws}`), making the
`(org, workspace)` → key map injective for arbitrary id bytes. The key is opaque,
internal, never wire-exposed.

## Two architecture findings (ADR-0088 amended in this PR)

1. **D4 crate-merge is INFEASIBLE.** Merging `credential-runtime` into
   `nebula-credential` was specced, but the facade depends on engine + storage
   (Exec) + tenancy (Business); `nebula-credential` is Core; `cargo deny` forbids
   Core→Exec. The facade stays a separate Exec crate (only the non-generic
   `Arc<dyn>` redesign survives, in place).
2. **D7 "single enforcement point" is half-true.** The *derivation* collapses to
   one function. *Enforcement* can't collapse to one physical `ScopeLayer`:
   tenancy is Business, runtime is Exec, Exec→Business is forbidden — same layer
   barrier. So api enforces via `ScopeLayer`, runtime enforces in-Exec via its
   facade owner checks.

Both recorded as ADR-0088 amendments.

## Scope / follow-ups

This PR is the derivation unification (the security-relevant core). Deferred:
deleting the dead `CredentialStoreHandle::Layered` manual-enforcement arm in api
(dead while the facade is unwired), and any single-physical-enforcer relocation
(needs its own ADR per the layer barrier).

## Validation

543 tests (api + credential-runtime + storage-port, nextest) + new injectivity
unit tests; clippy `-D warnings` (`--all-targets`); pre-push (clippy-full +
crate-diff-gate) green.

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Resolved potential credential owner ID format inconsistencies across system components by implementing a unified derivation method with improved collision-resistant encoding.

* **Refactor**
  * Standardized credential owner identification logic to use a single canonical method across all planes, reducing duplication and improving maintainability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3642,6 +3642,7 @@ dependencies = [
  "nebula-resilience",
  "nebula-schema",
  "nebula-storage",
+ "nebula-storage-port",
  "pretty_assertions",
  "serde",
  "serde_json",
--- a/crates/api/src/transport/credential.rs
+++ b/crates/api/src/transport/credential.rs
@@ -112,7 +112,10 @@ impl ScopeResolver for RequestCredentialOwner {
 }
 
 pub(crate) fn owner_id_from_scope(scope: &Scope) -> String {
-    format!("{}:{}", scope.org_id, scope.workspace_id)
+    // The single canonical derivation (ADR-0088 D7) — shared with the
+    // credential-runtime plane so both key the same tenant identically. Was a
+    // local `format!("{}:{}", …)` that drifted from the runtime's `/` form.
+    scope.credential_owner_id()
 }
 
 /// The credential store selection for CRUD operations.
--- a/crates/credential-runtime/Cargo.toml
+++ b/crates/credential-runtime/Cargo.toml
@@ -30,6 +30,9 @@ nebula-credential-testutil = { path = "../credential-testutil", optional = true 
 nebula-engine = { path = "../engine" }
 nebula-core = { path = "../core" }
 nebula-schema = { path = "../schema" }
+# Core `Scope` — the canonical credential owner_id derivation
+# (`Scope::credential_owner_id`, ADR-0088 D7) that `TenantScope` routes through.
+nebula-storage-port = { path = "../storage-port" }
 nebula-eventbus = { path = "../eventbus" }
 nebula-resilience = { path = "../resilience" }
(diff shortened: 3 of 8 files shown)


vanyastaff:
Fixed in `b0cca210` — you're right, `nebula-credential-runtime` does **not** depend on `nebula-tenancy` (only a doc-comment mentions `CredentialScopeLayer`). Dropped the bogus edge from the D4 amendment; the infeasibility conclusion is unchanged — the runtime's `nebula-engine` + `nebula-storage` (both Exec) deps alone make a merge into Core a forbidden Core→Exec inversion.

PR #776, opened 2026-06-04
feat(resource): async guard release + pool maintenance reaper + Manager split

vanyastaff:
## What

Four improvements to `nebula-resource`, derived from a cited deep-research pass on Rust 1.96 performance + code-style and verified with adversarial Rust reviews.

### 1. Explicit `async ResourceGuard::release(self) -> Result<(), Error>` (feat)
The awaited release checkpoint canon §11.4 calls for ("authors must not assume release ran without an explicit checkpoint"). Release callbacks now **return** the teardown future instead of submitting it: `Drop` submits-and-discards (best-effort, **byte-equivalent** behavior — unchanged), while `release()` awaits it inline and surfaces the recycle/destroy/reset `Result`. Drain accounting always settles (even on `Err`); `#384` permit-held-until-reset and bounded `S4` poison-on-failed-reset are preserved; the consumed guard is neutralized so a trailing `Drop` cannot double-settle.

### 2. Background pool maintenance reaper (feat)
`PoolRuntime::run_maintenance` (idle-timeout / max-lifetime / stale-fingerprint / revoked eviction) existed and was unit-tested but had **zero production callers** — so `idle_timeout` was effectively dead config. `Manager::register` now spawns a reaper per `Pooled` registration **only when a TTL is set** (zero overhead otherwise). It holds a `Weak<ManagedResource>` (never extends resource lifetime) + the manager cancel token: stops promptly on shutdown, self-terminates on `remove()`. New `ResourceEvent::MaintenanceEvicted` for observability.

### 3. `Manager` god-file split (refactor)
The single ~2100-line `impl Manager` block in `manager/mod.rs` (**2991 → 1149 LoC**) is carved into concern-named sibling `impl Manager` blocks: `acquire.rs`, `registration.rs`, `rotation.rs`. Pure move — public API unchanged (three private helpers elevated to `pub(crate)` for the in-mod test module).

### 4. `SlotCell::generation` ordering (perf)
Dropped a dangling `Acquire` on the no-entry fallback — it paired with no `Release` and synchronized nothing (the arc-swap `swap(None)` edge + per-location coherence already provide the guarantee). Now `Relaxed` + a one-line rationale.

## Verification
- `cargo clippy -p nebula-resource --all-targets -- -D warnings` — clean
- `cargo nextest run -p nebula-resource` — **343 passed** (incl. revoke-TOCTOU / S4 / #384 / shutdown-race + new release & reaper tests)
- `RUSTDOCFLAGS=-D warnings cargo doc -p nebula-resource --no-deps` — clean
- Adversarial Rust reviews on the split + both features: **no correctness defects**.

Rationale + the full prioritized action list: `crates/resource/docs/perf-style-research-2026-06.md` (added here).

## Notes
- The breaking `ResourceGuard` constructor signature change is contained to `crates/resource` (no external caller).
- `ResourceEvent` is `#[non_exhaustive]`, so the new variant is non-breaking.

## Release Notes

* **New Features**
  * Automatic background maintenance now evicts idle pooled instances based on configured timeouts.
  * ResourceGuard now provides an explicit `release()` method for controlled async cleanup.
  * New event type tracks pool maintenance evictions.

* **Documentation**
  * Added comprehensive performance and design research report.


vanyastaff:
Addressed all three review comments in `019f77e4`:

**[P1] `release()` cancellation (`guard.rs`, codex)** — Fixed. `release()` now runs its teardown **and** the drain settle on a detached `tokio::spawn`'d task and awaits the `JoinHandle`. Dropping a `JoinHandle` does not abort the task, so a caller aborted mid-teardown no longer drops the teardown future (the pooled runtime is still destroyed) nor skips the drain decrement (`graceful_shutdown` / `revoke_slot` no longer wedge on a stuck slot). A panicking resource teardown is caught (`catch_unwind`) and surfaced as a typed error while the drain still settles — parity with `Drop`. New regression test `release_teardown_survives_caller_cancellation` cancels `release()` mid-teardown (20ms timeout over a 300ms destroy) and asserts the destroy still ran and shutdown does not hang.

**`maintenance_interval` clamp (`registration.rs`, Copilot)** — Fixed. Floored at **1s** instead of 1ms, so a zero/tiny operator value degrades to a sane cadence rather than a CPU-burning spin loop.

**Flaky reaper test (`basic_integration.rs`, Copilot)** — Fixed. The test now polls for the `MaintenanceEvicted` **event** (draining the subscriber on every tick) instead of racing the destroy counter (which flips mid-sweep) against the later emit.

Verified: `clippy -D warnings` clean, **344** nextest pass, `rustdoc -D warnings` clean.

vanyastaff:
Addressed the CodeRabbit finding on `pool_maintenance_reaper_not_spawned_without_ttl` in `03fada9d`:

Correct catch — and a direct consequence of the 1s maintenance-interval floor added earlier in this PR. The test now (1) waits until the released instance is observed back in idle via `pool_stats` (so `destroy_counter == 0` can't false-pass on a never-completed recycle), and (2) observes `destroy_counter` for **1500ms** — longer than the 1s floor — so a spuriously-spawned reaper would have time to sweep and be caught.

Verified: `clippy -D warnings` clean, **344** nextest pass.

PR #788, opened 2026-06-10
refactor(credential): delete nebula-credential-testutil, dedup in-memory stores

vanyastaff:
## What

Deletes the `publish=false` `nebula-credential-testutil` crate (**7 → 6** credential crates). Its two in-memory store doubles (`InMemoryStore`, `InMemoryPendingStore`) duplicated the canonical `nebula_storage::credential` backends.

## Why it existed (and why the dup wasn't gratuitous)

`nebula-tenancy` (Business) cannot dep `nebula-storage` (Exec) — the layer rule enforced by the `deny.toml` wrappers — and an external dev-dep on storage breaks the `PendingStateStore` blanket bound via a two-copies cargo resolution. The shim crate was the workaround.

## How it's removed

- **`nebula-credential-runtime`** (Exec, already dev-deps storage with `credential-in-memory` + `test-util`) now uses `nebula_storage::credential::InMemoryPendingStore` directly (`service.rs`, `ops.rs`).
- **`nebula-tenancy`** keeps a colocated `#[cfg(test)]` `CredentialStore` double in `credential_scope::tests` — it only needs a store to wrap and still cannot reach the Exec adapter. This also sidesteps the two-copies bound by-construction and survives the upcoming `InMemoryStore` removal.
- Removed the crate, its workspace member, its `deny.toml` `[bans].deny` wrapper entry, and the `nebula-credential-testutil` entry from `nebula-credential`'s wrappers allowlist.
- Doc/comment sync: `storage/pending.rs` shim note, `credential` README + Cargo.toml, `MATURITY.md`, root `README.md`, two `.claude` skills, `.pi/conventions.json`.

## Evidence

- `cargo check --workspace` clean (api, server, examples)
- `nebula-tenancy` 32/32 · `nebula-credential-runtime` (`--features test-util`) 52/52 · `nebula-storage` (`--features test-util,credential-in-memory`) 275/275
- `clippy --all-targets -- -D warnings` clean on all three
- `cargo deny check` (advisories, bans, licenses, sources) ok
- lefthook pre-commit + pre-push green

## Follow-up (separate PRs)

- Remove `InMemoryStore` / `InMemoryPendingStore` entirely — credential store backends become **SQLite + Postgres only** (SQLite `:memory:` for tests).
- `#[derive(Credential)]` declarative authoring tier (own ADR + API-GATE).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
[...]

PR #789, opened 2026-06-10
feat(credential)!: durable SQLite + Postgres credential persistence (ADR-0088 D7)

vanyastaff:
## What

Replaces the **in-memory-only** credential store with real **durable SQLite + Postgres** backends, n8n-style (encrypted at rest in the DB, no external Vault). Credentials now survive a process restart — closing the gap that kept the active-credential-lifecycle moat latent.

Executes **ADR-0088 D7** (build durable backends on the live `StoredCredential` + `EncryptionLayer` path; delete the dead Model B row-model) with an **A-plus** amendment: a thin set of promoted identity columns (`name`/`owner_id`/`kind` + `UNIQUE(owner_id, name)`) so named/typed credentials are queryable without a future live-secret JSON→column migration.

> Stacked on #788 (testutil deletion). Base is that branch; this PR's diff is the durable-persistence work only.

## Phases (one commit each)

- **P0** `docs/plans/2026-06-10-durable-credential-persistence.md` — locked design.
- **P1** `StoredCredential.name: Option<String>` (API-GATE) + `SqliteCredentialStore` (refresh_claim adapter pattern: CAS via `UPDATE … WHERE version`, guarded `u64↔i64`, opaque-ciphertext `data` BLOB, `owner_id` from metadata, `#[tracing::instrument]`) + migration `0030_credentials_store.sql` (drops dead Model B tables) + shared conformance suite (InMemory + SQLite `mode=memory&cache=shared`).
- **P2** `PgCredentialStore` (native `TIMESTAMPTZ`/`BYTEA`/`BIGINT`, `ON CONFLICT … WHERE version`) + migration `0030` (postgres) + `DATABASE_URL`-gated PG conformance case.
- **P3** delete dead Model B Rust (`CredentialRepo`, `CredentialRow`/`CredentialAuditRow`/`PendingCredentialRow`) — zero impls, ADR-0088 D7 mandate.
- **P4** production `CredentialService` composes a durable `SqliteCredentialStore` by default (`NEBULA_CRED_DB`, default file `nebula-credentials.db?mode=rwc`; `sqlite::memory:` for ephemeral); `try_default_credential_service` is now `async`; shared body extracted to `with_store<S: CredentialStore>`. Pending-state store stays ephemeral in-memory (OAuth handshake, TTL ≤ 10 min, ADR-0084).

## Design notes

- `data` is an **opaque ciphertext BLOB** — the `EncryptionLayer` above the store serialises the AES-256-GCM envelope (carrying `key_id` for rotation), so no per-row key column is needed and the store never decrypts.
- Timestamps: SQLite stores millis-INTEGER (lexicographic-text ordering is fragile); Postgres uses native `TIMESTAMPTZ`. `put` returns the read-back row, so cross-backend `updated_at` equality holds without touching the in-memory reference store.
- Postgres is compile- + clippy-verified; its conformance case runs only with `DATABASE_URL` (no live DB in CI, same as the spec-16 PG adapters).

## Out of scope (follow-ups)

- Deleting the `InMemoryStore` *type* and running all tests on SQLite `:memory:` (it is now redundant; needs a focused PR — the test-helper async/Result conversion is broad).
- Retiring/relocating the unwired `nebula-credential-vault` crate.

## Evidence (bare)

`cargo check --workspace` clean · `nebula-storage` 347/347 (sqlite,postgres,test-util,credential-in-memory) · `nebula-credential-runtime` 52/52 (test-util) · `nebula-engine` 421/421 · `nebula-api` 490/490 · `nebula-credential`+`nebula-tenancy` 305/305 · `clippy --all-targets -- -D warnings` clean on all · `cargo deny check` ok · lefthook pre-commit + pre-push green.


vanyastaff:
Addressed all review findings in `67dbb608`:

- **🔴 P1 — durability bug (codex, `sqlite.rs`): fixed.** `connect()` ran migration `0030` (which starts with `DROP TABLE credentials`) on every call, so a restart against the same `NEBULA_CRED_DB` wiped the store. `connect()` now probes `sqlite_master` and applies the migration only when the `credentials` table is absent; an existing store keeps its rows. New regression test `sqlite_connect_is_idempotent_and_preserves_existing_rows` reconnects to the same DB and asserts the row survives.
- **UNIQUE-violation discrimination (`sqlite.rs`, `postgres.rs`): fixed.** `CreateOnly` no longer maps every UNIQUE violation to `AlreadyExists{id}`. SQLite inspects the failed-column message; Postgres branches on the constraint name (`credentials_pkey` → `AlreadyExists`, otherwise `Backend`). A `(owner_id, name)` collision on a new id now surfaces as an explicit `Backend` error.
- **CAS corruption masking (`sqlite.rs`, `postgres.rs`): fixed.** A negative stored `version` (table corruption) was masked as `VersionConflict { actual: u64::MAX }`; it now returns `Backend`.
- **Plan doc (`...durable-credential-persistence.md`): fixed.** Corrected the Postgres `metadata` column to `TEXT` (matching the migration + `PgCredentialStore`), not `JSONB`.

Evidence (bare): `nebula-storage` clippy `-D warnings` clean (`sqlite,postgres,test-util,credential-in-memory`); nextest **348/348**; `cargo check --workspace` clean.

PR #791, opened 2026-06-11
refactor(credential)!: consolidate the credential subsystem into one crate (ADR-0092)

vanyastaff:
## Summary

Collapses the entire credential subsystem into a single `nebula-credential` crate via **port-inversion**, per **ADR-0092**. The `credential-runtime → {engine, storage} → credential` cargo cycle that previously forced the split was proven (edge-by-edge) to be an **artifact of misplaced machinery**, not a fundamental dependency: `nebula-engine::credential/**` referenced zero Exec-only types. Relocating that machinery + inverting the one piece of real I/O (the OAuth2 refresh HTTP call) behind a narrow port dissolves the cycle.

**Two crates deleted** (`nebula-credential-runtime`, `nebula-credential-builtin`); their contents now live in `nebula-credential`. No cargo feature-gates were used for the split.

## Before → after

```
BEFORE                                  AFTER
nebula-credential        (contract)     nebula-credential  (contract + builtins + facade
nebula-credential-builtin (builtins)                        + resolver/refresh/lease/rotation/
nebula-credential-runtime (facade)                          dispatch + RefreshTransport port)
nebula-engine::credential (lifecycle)   nebula-engine      (reqwest-free; re-exports for API stability)
  ↑ cycle                               nebula-storage     (store adapters + Enc/Cache/Audit decorators)
                                        nebula-api         (secure-composition builder +
                                                            ReqwestRefreshTransport = composition root)
```

- `nebula-credential` depends only **downward** (`storage-port`, `crypto`, `core`/`schema`/`metadata`/`error` + cross-cutting). **No `nebula-engine` / `nebula-storage` edge.**
- The decorators (`Encryption`/`Cache`/`Audit`) stay in `nebula-storage` (they are storage-coupled); the `AuditSink` **trait** + audit value types moved to Core so the refresh coordinator can emit without an upward edge.
- `nebula-engine` is now **reqwest-free** — `ReqwestRefreshTransport` lives at the api composition root.

## Security seams (deliberately preserved)

- **`RefreshTransport`** (new port): drawn **narrow** = `post_token(TokenPostRequest) -> TokenPostResponse{status, body}`. SSRF endpoint validation (https-only, no localhost/private/link-local), OAuth2 form composition, status interpretation, body parse, and SEC-02 redaction **all stay in `nebula-credential`, before/around the transport call**. The reqwest impl is a dumb bounded-POST pipe — zero validation, zero parsing. A wider seam would have exported the SSRF defense to a second composition root.
- **`StaticKeyProvider`** (static encryption key, dangerous in release): re-gated `#[cfg(test)]` (never compiled into a release binary, ADR-0023). The `#281` silent-decrypt / `#282` lazy-rotation / SEC-11 AAD-mandatory encryption tests are **preserved**.
- **Secure construction**: the `Audit(Cache(Encryption(raw)))` composition is preserved verbatim in the api-tier builder; `CredentialService::from_secure_parts` is the single composition-root constructor.

## Breaking changes

- `nebula-credential-runtime` and `nebula-credential-builtin` deleted — import the facade (`CredentialService`, `CredentialServiceError`, …) and the built-ins (`BearerTokenCredential`, `SharedKeyCredential`, `SigningKeyCredential`, `register_builtins`) from `nebula_credential`.
- `RefreshTransport` must be injected at the composition root (`nebula_api::ports::ReqwestRefreshTransport`).
- The nebula `test-util` cargo feature was eradicated from `nebula-credential` and `nebula-storage` (tokio's `test-util` is unaffected).
- `nebula-engine` no longer depends on `reqwest`.

## Tests deprioritized during the move (owner-directed)

Test modules that depended on in-memory doubles / now-`#[cfg(test)]` cross-crate helpers were dropped to be rewritten against the consolidated crate in a follow-up. Production code stays green; security-regression coverage (`#281`/`#282`/SEC-11, conformance, tenant-isolation) was **kept** via local helpers, not dropped.

## Verification

`cargo clippy --workspace --all-targets -- -D warnings` clean · `cargo deny check` clean · per-crate nextest: credential **341** · storage **280** · api **490** · engine **327** · resource green. `cargo tree` confirms no `credential → engine/storage` edge and no `sqlx` in `action`/`plugin`/`sdk`.

## Follow-ups (not in this PR)

- Gate `nebula-metrics`' OTLP-HTTP exporter so the refresh coordinator's metric-handle dependency stops transitively dragging `reqwest` into `action`/`plugin`/`sdk` trees (pre-existing, separate crate).
- Refresh stale per-crate `CLAUDE.md` lines that still say "runtime credential resolution lives in `nebula-engine::credential`".

## Summary by CodeRabbit

* **Architecture**
  * Consolidated credential subsystem into a single crate for simpler integration.

* **New Features**
  * Two-tier refresh coordination with improved crash detection and reclamation.
  * Pluggable, bounded OAuth2 token transport for safer token refreshes.
  * Durable audit events for credential operations and escalation alerts.
  * Built-in credential types registration and new crypto adapter traits.

* **Chores**
  * Removed legacy test utilities and cleaned up obsolete fixtures/tests.


vanyastaff:
## Review feedback triage (commit `8c7c3d4f`)

**Fixed:**
- **builtin resolve tests weakened** (`bearer_token`/`shared_key`/`signing_key`) — the 6a fold had switched exact-secret `assert_eq!` to non-empty checks. Restored exact expected-value assertions (`sk-abc123` / `psk-xyz` / `whsec_1`). Real regression — thanks.
- **`register_builtins` doc overclaim** — clarified it registers the three *reference* credentials; `api_key`/`basic_auth`/`oauth2` are intentionally not auto-registered.
- **`ReqwestRefreshTransport` silent fallback** — now `warn!`s when the hardened client-builder fails so the lost connect-timeout / redirect-cap is observable.
- **`from_secure_parts` accessibility** — marked `#[doc(hidden)]` with a rationale comment.

**Won't-fix (verified invalid against current code):**
- **`rand::random_range` "Critical"** — compiles and CI is green; the free-fn form is valid in the pinned `rand`. Not a build break.
- **`l1.rs:215` `unwrap()` "Major"** — it's a `const` (`NonZeroUsize::new(CONST).unwrap()`), which is explicitly allowed (tests/const/bin exemption) and the doc proves infallibility. The `unreachable!()`s are `// guard-justified:` per repo policy.
- **`token_refresh` client_id "secret leak"** — `client_id` is the OAuth2 Basic-auth *username* and is non-secret (public per RFC 6749); `SecretString` is correctly applied to `client_secret` only.
- **`sentinel.rs:148` duration fallback** — `chrono::Duration::from_std` only fails past ~292 billion years; the input is a bounded config value, so the fallback is unreachable in practice.

**Deferred (pre-existing, out of scope for this refactor):**
- **`AuditResult::Error(String)` secret-safety** — valid hardening idea, but this is a pre-existing type (relocated, not introduced here) and the suggested change is a type redesign touching all audit producers. Tracked as a follow-up rather than expanding this consolidation PR.

`from_secure_parts` residual: `nebula-credential` is unpublished and the `deny.toml` wrappers allowlist restricts its dependents to trusted in-workspace composition roots; `nebula-sdk` (the only public crate) never re-exports it — so there is no external caller that could bypass the builder.

PR #857, opened 2026-06-19
feat(plugin-core): add core.aggregate action (array → grouped/scalar summary)

vanyastaff:
## What

Adds `core.aggregate` — reduce a JSON array of objects to summary rows: `count` / `count_distinct` / `sum` / `avg` / `min` / `max` / `collect` / `join`, with an optional `group_by`. Pure StatelessAction, same template as `core.filter`.

## Why (non-redundant)

The `{{ }}` expression language has **no live array aggregator** (no `sum`/`avg`/`count`/`min`/`max` over an array), and `group_by`/`reduce` are lambda-gated — unreachable from non-lambda config. So aggregation is only available as a node. (This is the array-predicate gap the lambda-less expression language can't express; `object.pick`/`omit`/`merge` already cover single-object reshape inline, which is why `core.map` is held pending an architecture decision.)

## Contract (adversarially reviewed)

- `data` null/absent/non-array → **Fatal**; every element must be a JSON object (`is_object()` guard); empty `aggregations` or duplicate `out` key → Fatal; missing `group_by` key on an element → Fatal.
- **`on_error` honesty:** numeric aggs (`sum`/`avg`/`min`/`max`) treat a **missing / null / non-numeric** value per `on_error` — `Fail` (default) → Fatal naming the field (so `fail` genuinely catches dirty data instead of silently shrinking the `avg`/`sum` denominator); `Skip` → ignore that value. `count(*)` always counts the row; `count_distinct`/`collect`/`join` skip null/missing.
- **Integer-preserving `sum`** (i64 `checked_add`, overflow → Fatal; any float → f64); `avg` always f64; `min`/`max` return the actual element value. Float `sum`/`avg` that overflow to non-finite → **Fatal** (no silent zero).
- Output is always a `Value::Array` of rows (length 1 ungrouped) in **deterministic first-seen group order**; empty input → one zeroed row (ungrouped) / `[]` (grouped).

## Tests

25 unit tests (Fatal witnesses; `on_error` honesty incl. avg/min/max-null-under-Fail + skip-denominator; integer preservation; float-overflow→Fatal; `count_distinct` type distinction `1`/`"1"`/`1.0`; multi-key `group_by`) + a plugin-wiring e2e (RED no-plugin → GREEN groups+sums and reaches `Completed`). All assert values.

Catalog now: `core.{set_fields, json_transform, if, switch, datetime, filter, aggregate}` + combinators.

## Release Notes

* **New Features**
  * Introduced `core.aggregate` action for grouping and summarizing JSON array data
  * Supports aggregation functions: count, count_distinct, sum, avg, min, max, collect, and join
  * Configurable error handling with strict input validation

* **Tests**
  * Added end-to-end tests validating aggregate action behavior

### DIFF
--- a/crates/plugin-core/src/actions/aggregate.rs
+++ b/crates/plugin-core/src/actions/aggregate.rs
@@ -1,0 +1,1484 @@
+//! `core.aggregate` — reduce a JSON array of objects to grouped summaries.
+//!
+//! Iterates the input array, optionally partitions elements into groups by
+//! one or more field keys, and computes one or more aggregation functions
+//! per group. The output is always a JSON array of summary objects — one row
+//! per group (or one row when no grouping is requested).
+//!
+//! This fills the aggregation gap in the `{{ }}` expression language, which
+//! has no live array aggregators (`sum`/`avg`/`count`/`min`/`max` over an
+//! array) and whose `group_by`/`reduce` builtins require lambda support that
+//! is not yet implemented.
+//!
+//! ## Input
+//!
+//! ```json
+//! {
+//!   "data": [
+//!     { "region": "west", "amount": 10 },
+//!     { "region": "east", "amount": 20 },
+//!     { "region": "west", "amount": 30 }
+//!   ],
+//!   "group_by":     ["region"],
+//!   "aggregations": [
+//!     { "fn": "count", "out": "n" },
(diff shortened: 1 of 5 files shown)


vanyastaff:
@coderabbitai review

PR #865, opened 2026-06-21
feat(storage): mint-on-park resume-token store — ADR-0099 W-S3c

vanyastaff:
## Summary

- **Port contract** (`nebula-storage-port`): `ResumeTokenRow` DTO, `TokenHash` 32-byte newtype with `subtle::ConstantTimeEq`-backed equality (no hex/Display), `ResumeTokenWaitKind` enum, `ResumeTokenStore` trait (`consume` atomic-delete-and-return, `revoke_on_terminal` count-returning sweep), and `TransitionBatch::resume_tokens` field so token creation is never split from the state snapshot.
- **InMemory backend**: `InMemoryResumeTokenStore` shares `SharedState` mutex with `InMemoryExecutionStore`; `pub(super) new` / `pub standalone()` split; timing-safe `ConstantTimeEq` in `consume`.
- **SQLite backend**: `DELETE … RETURNING *` single-use atomic delete; `ON CONFLICT(execution_id, node_key) DO NOTHING` idempotent re-park guard; FK corrected to simple `REFERENCES port_executions (id)` (SQLite FK must reference PK).
- **Postgres backend**: same RETURNING + conflict clause; compile-verified, runtime-gated on `DATABASE_URL`; migration `0035_port_resume_tokens.sql`.
- **Engine mint-on-park**: `mint_park_token` — `rand::rng().fill_bytes` → SHA-256 → `SecretString` bearer; plaintext never touches `String`/`format!`/log/trace/`Debug`; `revoke_on_terminal` called on every terminal transition.
- **11 tests** (all RED-first): 8 InMemory, 2 SQLite (`#![cfg(feature="sqlite")]`), 1 engine integration (webhook park → token visible in store).

## Security constraints preserved

- Plaintext token never leaves `SecretString` / raw byte array — no `String`, no `format!`, no `Display`, no `Debug` derive on the cleartext.
- `consume` has no `scope` parameter — scope comes FROM the returned row (no confused-deputy bypass).
- No non-consuming `resolve` method (replay footgun).
- No `mint` method — minting is done exclusively inside the `TransitionBatch` commit path.
- Token hash is raw 32-byte SHA-256 stored as `BLOB`/`Vec<u8>`, never hex text.
- Never logged or traced at any level.

## Test plan

- [x] `cargo nextest run -p nebula-storage --test resume_token_inmem` — 8/8 pass
- [x] `cargo nextest run -p nebula-storage --test resume_token_sqlite --features sqlite` — 2/2 pass
- [x] `cargo nextest run -p nebula-engine --test resume_tokens` — 1/1 pass
- [x] `cargo clippy -p nebula-storage -p nebula-engine --all-targets --all-features -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `RUSTDOCFLAGS=-D warnings cargo doc --no-deps -p nebula-storage -p nebula-engine` — clean
- [x] Lefthook pre-push (clippy-full + crate-diff-gate) — all gates green

## Summary by CodeRabbit

* **New Features**
  * Added resume token support for workflows. When executions pause waiting for external signals (webhooks or approvals), secure single-use resume tokens are now automatically generated and persisted, enabling safe external reconnection and execution resumption.


vanyastaff:
Both Codex P1s are addressed in `a9df4aeb` (verified against source + 950 tests green):
- **PG FK → unique key**: the composite FK was changed to single-col `REFERENCES port_executions (id) ON DELETE CASCADE` across both `0035` migrations + `postgres/schema.sql` + `sqlite/schema.sql` (the real PK; closes SQLSTATE 42830 on `init_schema`).
- **rebind drops resume_tokens**: `ScopedExecutionStore::rebind` now carries + re-scopes `resume_tokens` to the bound tenant (alongside outbox), with a RED→GREEN cross-tenant-denial test; a `ScopedResumeTokenStore` decorator covers the read side.
Plus the full 4-layer review fixes (in-mem PK parity, `wait_kind` CHECK, preimage zeroize, `expires_at` from `wake_at`, `#[non_exhaustive]`, object-safety probe). `revoke_on_terminal` wiring is intentionally W-S3e per ADR-0099; doc made honest.

PR #866, opened 2026-06-21
feat(api): W-S3d — webhook→Resume producer (attacker-facing) (ADR-0099)

vanyastaff:
## What

The final wait-state slice (ADR-0099 W-S3d): the **inbound transport** that makes the resume machinery reachable. `POST /resume` accepts a bearer resume token, consumes it against the W-S3c store (single-use), and enqueues a **targeted** `ControlCommand::Resume`. This is the only attacker-reachable wait-state surface — **security-auditor RELEASE-GATE**.

## Security design (born already-safe)

- **Bearer in `Authorization: Bearer` header, never the URL** — no secret-in-logs.
- **Mounted pre-tenancy** — the handler has **no** `TenantContext` extractor, so scope-from-row is enforced by *absence*: `scope`/`execution_id`/`target` all come from the resolved `ResumeTokenRow`, never the request.
- **Anti-guessing rate-limit = per-IP + global tiers, BEFORE any store hit** (per-token limiting is worthless against rotation; the global tier is the IP-trust-independent backstop). Per-tenant tier runs post-consume.
- **Single-use `consume`** (`DELETE…RETURNING`) → **kind-match fail-closed** (Webhook only; `#[non_exhaustive]` `_`→404) → **expiry** via injectable `Clock` (fail-closed on parse failure) → enqueue via `ScopedControlQueue` bound to `row.scope`.
- **Uniform 404** for absent/expired/consumed/forged/malformed-bearer (no existence oracle).
- **Storage error → 503 + Retry-After**, token unconsumed (a transient DB blip must not silently drop a legitimate resume).
- **Body inert** (never deserialized into the enqueue); HMAC-gated body mode hard-blocked. Bearer dropped immediately after hashing; never logged.

## Tests

14 integration tests (`crates/api/tests/resume_e2e.rs`): happy path, single-use replay, forged/absent, expired, malformed-expiry, kind-confusion, scope-from-row, per-IP/global/per-tenant rate-limit ordering, no-token-in-URL, storage-err→503, body-inert, bearer-extraction uniformity. clippy `--all-features -D` clean; 554 nebula-api tests pass; zero library-path `expect`/`unwrap`/`panic!`.

## Abuse-case coverage

15-case enumeration (chief-architect plan): token guessing/rotation/DoS, replay, cross-tenant, confused-deputy, kind-confusion, rate-limit bypass, timing/existence oracle (rate-limited residual), enqueue-without-consume, body-trust, expired-resume, secret-in-logs, storage-error-downgrade. The timing-oracle residual and the per-replica rate-limit cap are flagged for the RELEASE-GATE.

## Summary by CodeRabbit

* **New Features**
  * Added a new `POST /resume` webhook callback with Bearer token handling and a request body size limit.
  * Enforced fail-closed validation, single-use replay protection, and staged rate limiting (per-IP, global, per-tenant).
  * Returns `202` on success, `404` for invalid/expired tokens, `429` when rate-limited, and `503` for transient storage issues (with `Retry-After` where applicable).
* **Infrastructure**
  * Wired resume storage/producer components across Memory/SQLite/Postgres so resume operations use the same durable backend state.
* **Tests**
  * Added expanded end-to-end and producer contract tests, including SQLite atomicity/durability cases.

### DIFF
--- a/apps/server/src/compose.rs
+++ b/apps/server/src/compose.rs
@@ -225,6 +225,13 @@ pub(crate) struct ExecutionStoreBundle {
     /// in-memory store on a durable backend silently breaks the round-trip:
     /// the engine writes to the pool; the producer reads from a different empty store.
     resume_token_store: Arc<dyn nebula_storage_port::store::ResumeTokenStore>,
+    /// Resume producer (W-S3d Option B1) — the atomic consume+enqueue seam the
+    /// `POST /resume` handler uses. MUST share the SAME backend pool / shared
+    /// state as `execution_store` and `control_queue` so the token DELETE and the
+    /// `Resume` INSERT commit in one transaction. A mismatched pool would split
+    /// the burn and the enqueue across two backends — exactly the durability gap
+    /// this seam closes.
+    resume_producer: Arc<dyn nebula_storage_port::store::ResumeProducer>,
 }
 
 /// Build the execution-store bundle for the configured backend.
@@ -284,6 +291,9 @@ fn build_memory_execution_stores() -> Result<ExecutionStoreBundle, TransportInit
     // Resume-token store shares the same SharedState as the exec_store so that
     // tokens committed via TransitionBatch are visible to the POST /resume handler.
     let resume_token_store = exec_store.resume_token_store();
+    // Resume producer shares the SAME SharedState (token map + control queue) so
+    // the consume + enqueue happen under one lock.
+    let resume_producer = exec_store.resume_producer();
     let node_results = InMemoryNodeResultStore::new();
     let workflow_versions = InMemoryWorkflowVersionStore::new();
     let workflow_store = InMemoryWorkflowStore::new_with_versions(&workflow_versions);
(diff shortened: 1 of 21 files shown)


vanyastaff:
**Resolution of the two recurring bot findings (by-design / already-addressed — not blocking):**

1. **`app.rs` `/resume` not in the OpenAPI router (CodeRabbit, recurring "Heavy lift"):** intentional. `/resume` is an attacker-facing, pre-tenancy route deliberately mounted via plain `Router::merge` — the *same* established escape-hatch the webhook transport routes (`app.rs:132`) and the `/internal/v1/*` routes (`app.rs:161`) use to stay out of `/api/v1/openapi.json`. The exemption is documented inline (`app.rs:137-146`). Not a public documented API; no spec drift to detect.

2. **`resume.rs` `Bytes` buffers the body before the in-handler cap (CodeRabbit + Codex P2, recurring):** the pre-buffer bound is enforced by the `DefaultBodyLimit::max(RESUME_BODY_LIMIT_BYTES)` tower layer on the `/resume` `MethodRouter` (`app.rs:152`) — axum rejects an oversized body at the extractor (413) before the handler runs. The in-handler `body.len()` check is defense-in-depth only. Verified by `body_inert_extra_json_ignored_oversized_body_413` (5 KiB → 413 through the full tower stack via `oneshot`). The bots flag this because they analyse `resume.rs` in isolation and don't see the layer in `app.rs`.

The substantive Codex P1 (token burned before the Resume is durably enqueued → permanent `Paused` hang) is closed by the atomic `ResumeProducer` (consume+enqueue in one transaction); the CodeRabbit Major on the debug-only command assert is closed by the release-enforced pre-mutation guard. Security RELEASE-GATE re-signed-off on the atomic seam.

PR #984, opened 2026-08-05
feat(storage-port/storage)!: end dispatch claims at durable runtime handoff (#976)

### LINKED ISSUE #976: feat(engine/storage)!: end dispatch claims at durable runtime handoff
## Goal

Close NS05: action duration must not extend dispatch-queue claim ownership.

## Architecture

- A dispatch claim protects delivery only.
- Before acknowledging the queue row, runtime control durably accepts ownership of the execution turn under the aggregate lease/fence.
- After handoff, long-running work is governed by execution lease and persisted recovery state, not by extending the dispatch claim.
- Crash before handoff acknowledgement redelivers; crash after acknowledgement recovers from durable aggregate truth.

## Scope

- Define a typed durable handoff command/outcome at the runtime-control storage boundary.
- Refactor processor flow to claim -> exact validation -> durable handoff -> queue ack -> execute/recover.
- Fence all handoff and acknowledgement writes with storage-minted generations.
- Add bounded shutdown behavior and telemetry for each phase.

## Acceptance

- A deliberately blocked action outlives the dispatch claim without claim renewal or duplicate live owners.
- Crashes at every boundary converge on one fenced runtime owner.
- Stale claim generations cannot acknowledge or mutate a reclaimed row.
- InMemory, SQLite, live PostgreSQL, chaos, and focused loom/concurrency evidence pass.
- `dispatch_claim_extended_by_action_duration == false`.

vanyastaff:
Closes #976.

## The problem

`Orchestrator::handle_entry` awaits `sink.dispatch(&msg)` — a whole execution step that, by its own comment, "can park on a durable timer or an external wait" — and only then calls `mark_dispatched`. The dispatch claim is therefore held for the entire action duration, which makes the queue's reclaim timeout an implicit limit on how long an action may run. A slow action outlives its claim, a sweep redelivers the row, and a second worker starts a turn the first is still running.

Renewing the claim while the action runs does not fix it: it turns the queue row into a second, weaker lease competing with the execution aggregate's own.

## The shape

`ExecutionTurnHandoff::accept_turn` ends the claim at a definite point. It acquires the execution lease under the aggregate fence **and** acknowledges the queue row in one transaction, so the two can never disagree:

- Crash before the commit — the row is still `Processing`, the sweep redelivers, and no turn was accepted.
- Crash after — the row is acknowledged and the execution holds a durable lease, so recovery drives from aggregate truth rather than from the queue.

The trait doc explicitly forbids assembling this from `acquire_lease` + `mark_dispatched`: those are separate operations, and the crash window between them is the state this capability exists to make unreachable.

`TurnHeldByAnotherOwner` deliberately leaves the queue row unacknowledged. Acknowledging there would drop the turn on the floor — the row would be terminal while no owner ever ran it — so it stays claimable and the sweep redelivers once the competing lease expires.

The handoff carries a `JobClaimToken`, not a processor identity: a stable processor may claim a row, lose it to a sweep, and claim it again, so an acknowledgement issued against the first claim would terminalise a row the second is still working.

## Backends

SQLite takes the write lock up front with `BEGIN IMMEDIATE`. PostgreSQL re-reads the claim row `FOR UPDATE`, closing the window where a reclaim sweep could move it between the check and the acknowledgement. Both stamp the lease deadline in the database, matching `acquire_lease` — a caller's clock never decides liveness, because one running fast would fence out a healthy peer and one running slow would mint an already-dead lease, neither detectable from the row afterwards.

When the lease cannot be taken, one extra read distinguishes a live foreign lease from a missing execution: a caller must not treat a nonexistent execution as contention it can wait out.

## Two findings from writing the evidence

Both recorded rather than smoothed over:

- **`SELECT 1` is INT4 in PostgreSQL, not INT8.** Decoding the existence probes as `i64` failed at the driver. Only the live-PostgreSQL run could surface this; a SQLite-only suite would have shipped it.
- **The reclaim predicates disagree at the boundary.** The in-memory model uses `elapsed >= reclaim_after`, SQLite uses `processed_at_ms < cutoff`, so a zero window reclaims immediately on one backend and never on the other. The SQL cases backdate the claim explicitly rather than depend on either reading. The divergence is pre-existing reclaim behaviour, named here so it is not rediscovered later as a flake.

## Scope note

This PR lands the storage capability and its three-backend evidence. The orchestrator refactor to `claim → validate → handoff → dispatch` follows on this branch; the acceptance criterion about a blocked action outliving its claim is proven end-to-end there.

## Evidence

- 6771 workspace tests pass (3 skipped); 598 nebula-storage tests.
- 4 handoff cases pass on each of in-memory, SQLite, and live PostgreSQL 17-alpine.
- The PostgreSQL suite is added to the required `postgres-conformance` job rather than left to skip.
- clippy clean on default, sqlite-only, postgres-only, and all-features; `task quality` green.

## Summary by CodeRabbit

* **New Features**
  * Added reliable dispatch-claim handoff to execution turns across in-memory, PostgreSQL, and SQLite storage.
  * Execution leases and queue acknowledgments now complete atomically before dispatch.
  * Added fenced execution resumption and lease-based recovery after dispatch failures.
  * Added handoff outcome metrics and configurable handoff lease durations.
  * Added handling for superseded claims, conflicting leases, missing executions, and cross-tenant requests.

* **Tests**
  * Added conformance and integration coverage across storage backends.

### DIFF
--- a/.github/workflows/test-matrix.yml
+++ b/.github/workflows/test-matrix.yml
@@ -294,7 +294,8 @@ jobs:
             refresh_claim_pg_integration \
             revision_catalog_conformance_postgres \
             revision_catalog_migration_postgres \
-            schema_setup_postgres
+            schema_setup_postgres \
+            turn_handoff_conformance_postgres
           do
             cargo nextest run "${common[@]}" --test "$suite"
           done
--- a/apps/server/src/compose.rs
+++ b/apps/server/src/compose.rs
@@ -268,6 +268,10 @@ pub(crate) struct WorkerStoreProjection {
     /// different backends is exactly the miswiring that leaves an accepted
     /// execution parked forever with every component reporting healthy.
     pub(crate) control_queue: Arc<dyn nebula_storage_port::store::ControlQueue>,
+    /// Handoff over the **same** backend the queue and execution store use:
+    /// it commits the lease write and the queue acknowledgement in one
+    /// transaction (#976), so it must share their boundary.
+    pub(crate) turn_handoff: Arc<dyn nebula_storage_port::store::ExecutionTurnHandoff>,
 }
 
 #[cfg(feature = "runtime-repair-red")]
@@ -409,6 +413,7 @@ fn build_memory_execution_stores() -> Result<ExecutionStoreBundle, TransportInit
             },
             job_dispatch_queue: Arc::new(InMemoryJobDispatchQueue::new(&exec_store)),
             control_queue: Arc::clone(&shared_control_queue),
+            turn_handoff: Arc::new(nebula_storage::inmem::InMemoryTurnHandoff::new(&exec_store)),
         }
     };
 
@@ -451,7 +456,9 @@ async fn build_sqlite_execution_stores(
         SqliteWorkflowStore, SqliteWorkflowVersionStore, init_schema,
     };
     #[cfg(feature = "runtime-repair-red")]
-    use nebula_storage::sqlite::{SqliteIdempotencyGuard, SqliteJobDispatchQueue};
+    use nebula_storage::sqlite::{
+        SqliteIdempotencyGuard, SqliteJobDispatchQueue, SqliteTurnHandoff,
+    };
     use sqlx::sqlite::{
         SqliteConnectOptions, SqliteJournalMode, SqlitePoolOptions, SqliteSynchronous,
     };
@@ -525,6 +532,8 @@ async fn build_sqlite_execution_stores(
         },
         job_dispatch_queue: Arc::new(SqliteJobDispatchQueue::new(pool.clone())),
         control_queue: Arc::clone(&shared_control_queue),
+        // Same pool as the queue and execution store — see the field docs.
+        turn_handoff: Arc::new(SqliteTurnHandoff::new(pool.clone())),
     };
     #[cfg(feature = "runtime-repair-red")]
     let backend_lifecycle = ProfileBackendLifecycle::Sqlite(pool.clone());
@@ -572,7 +581,7 @@ async fn build_pg_execution_stores(
         init_schema as pg_init_schema,
     };
     #[cfg(feature = "runtime-repair-red")]
-    use nebula_storage::postgres::{PgIdempotencyGuard, PgJobDispatchQueue};
+    use nebula_storage::postgres::{PgIdempotencyGuard, PgJobDispatchQueue, PgTurnHandoff};
     use sqlx::postgres::PgPoolOptions;
 
     let environment_dsn;
@@ -645,6 +654,8 @@ async fn build_pg_execution_stores(
         },
         job_dispatch_queue: Arc::new(PgJobDispatchQueue::new(pool.clone())),
         control_queue: Arc::clone(&shared_control_queue),
+        // Same pool as the queue and execution store — see the field docs.
+        turn_handoff: Arc::new(PgTurnHandoff::new(pool.clone())),
     };
     #[cfg(feature = "runtime-repair-red")]
     let backend_lifecycle = ProfileBackendLifecycle::Postgres(pool.clone());
--- a/apps/server/src/runtime_repair_red/profile.rs
+++ b/apps/server/src/runtime_repair_red/profile.rs
@@ -256,6 +256,7 @@ impl RuntimeRepairHarness {
                 worker_projection.execution_stores,
                 worker_projection.workflow_stores,
                 worker_projection.job_dispatch_queue,
+                worker_projection.turn_handoff,
                 PROFILE_PROCESSOR_ID,
                 engine_clock,
                 execution_event_bus,
--- a/apps/worker/src/compose.rs
+++ b/apps/worker/src/compose.rs
@@ -14,7 +14,7 @@ use nebula_engine::{
 use nebula_metrics::MetricsRegistry;
 use nebula_plugin::{ManifestError, PluginError};
 use nebula_plugin_core::CorePlugin;
-use nebula_storage_port::store::JobDispatchQueue;
+use nebula_storage_port::store::{ExecutionTurnHandoff, JobDispatchQueue};
 use nebula_worker::{WorkerBuildError, WorkerRuntimeBuilder};
 
 /// Typed errors emitted by the composition root.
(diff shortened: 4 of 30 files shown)


vanyastaff:
## Orchestrator wiring: attempted, and it surfaced a design conflict worth naming

I wired `claim → handoff → dispatch` end to end (orchestrator, worker builder, both composition roots, and every test call site). It compiles, and the orchestrator's own suite goes green — including a rewritten NS05 case proving a blocked action no longer loses its row to the reclaim sweep.

Then four integration tests failed, and the cause is not the wiring:

**`WorkflowEngine` acquires the execution lease itself.** `crates/engine/src/engine/resume.rs` calls `acquire_lease` at four sites. Once the handoff holds a live lease, those calls return `Ok(None)` and the engine cannot drive the execution, so `worker_runtime_drives_execution_to_completed`, `reclaim_then_rerun_drives_exactly_once_via_real_sink`, `trigger_dispatch_end_to_end_real_engine_resume`, and `core_flavor_runtime_processes_seeded_start_job` all time out.

That is exactly the "second, weaker lease" this issue set out to remove, one layer further in: the turn is currently acquired twice, once by whoever accepts it and once by the engine. Making the handoff authoritative means `ExecutionSink::dispatch` receiving the `FencingToken` from the handoff instead of minting its own, and the engine's resume entry points taking a fence rather than acquiring one.

That is a real change to the sink contract and the engine's resume surface, not a mechanical follow-up, so I have not bundled it here. This PR remains the storage capability plus its three-backend evidence, which stands on its own and is what the acceptance criteria about fencing, tenant isolation, and crash boundaries are proven against.

The wiring is saved and reapplies cleanly (979 lines) for whoever takes the engine-side change; the branch is green without it.

**Remaining for this issue after the engine change:** the `dispatch_claim_extended_by_action_duration == false` assertion end to end, plus the chaos and loom evidence the acceptance list asks for.

vanyastaff:
## Orchestrator refactor — `claim → validate → handoff → dispatch` (26f02828)

The storage capability now has its production consumer. `Orchestrator::handle_entry` no longer holds a claim across the action: it ends the claim at `accept_turn` **before** the sink runs, and never acknowledges queue state after the handoff commits.

### The shape

- **validate** — the execution id must parse before anything is written; garbage rows are terminalised (`mark_failed`) instead of redelivering until exhaustion.
- **handoff** — `accept_turn` with the claim token: `Accepted{fence}` → dispatch; `ClaimSuperseded` → walk away; `TurnHeldByAnotherOwner` → leave the row claimable for sweep redelivery; `NotFound` → orphaned dispatch, `mark_failed`; backend error → leave `Processing` for reclaim.
- **dispatch** — `ExecutionSink::dispatch(&DispatchedTurn{msg, fence})`: the sink drives the turn **under** the handoff fence. `EngineExecutionSink` calls the new `WorkflowEngine::resume_execution_leased`, which adopts the fence (same heartbeat loop, fence-gated renew/release) instead of acquiring — a live lease blocks acquisition outright, even for the same holder, so acquiring could never succeed.
- Post-handoff failures record `failed` for operators and touch no queue state: recovery drives from the execution lease and persisted aggregate truth.

### Wiring

- `Orchestrator::new` requires `Arc<dyn ExecutionTurnHandoff>`; `WorkerRuntimeBuilder` gains `with_turn_handoff` + `WorkerBuildError::NoTurnHandoff` (fail-closed at build time — no silent return to pre-#976 behaviour). `with_handoff_lease_ttl` override (default 30 s, matching `DEFAULT_EXECUTION_LEASE_TTL`).
- apps/worker (SQLite + Postgres + in-memory) and apps/server runtime-repair projection construct the handoff over the **same** pool/core as the queue and execution store; the runtime-repair compose signature carries it through.
- New counter `nebula_orchestrator_handoff_total{outcome=accepted|claim_superseded|turn_held|error}` — vocabulary mirrors the storage crate's `acceptance_label`; dispatch counter semantics updated accordingly.

### Evidence

- **NS05 acceptance** (`blocked_action_outlives_claim_without_renewal_or_duplicates`): sink blocked 6× past `reclaim_after`; reclaim sweep finds `reclaimed == 0`, a second processor claims nothing, `acquire_lease` probe returns `None` (no duplicate live owner), exactly one sink entry; the execution row carries the orchestrator's lease **before** the sink returns. Red-able: reverting to dispatch-first restores the reclaimable-while-running row and both assertions fail.
- New unit coverage: invalid execution id → pre-handoff `mark_failed`; orphaned execution → handoff `NotFound` terminalises; post-handoff sink failure leaves the row terminal with the lease held for recovery.
- Existing suites re-proven under the new flow: routing, terminal-not-reclaimed, pre-handoff crash reclaim (RECLAIMED/EXHAUSTED), graceful single-row and multi-row batch flush.
- 559 tests green across engine/orchestrator/worker/metrics; 19 worker-bin tests incl. `runtime-repair-red`; storage handoff conformance unchanged and green; `cargo check --workspace --all-targets` clean; clippy `-D warnings` clean; fmt clean.
- Live-PostgreSQL orchestrator evidence rides the same trait surface proven by the storage conformance suite on this branch; the runtime-repair scenarios binary compiles under `runtime-repair-red,postgres` (runs where CI provides the database).

vanyastaff:
## Review findings addressed (289d64f5, 5a89e76a, 217a1343)

**🟠 Major — adopt path broke `fail_cold_start_preflight`'s durable failure write** (`resume.rs`) — confirmed valid, fixed in 289d64f5. The pre-flight failure path now threads the handoff fence (`ResumeLeaseSource::Adopt` → `handoff_fence: Some(fence)`): `fail_cold_start_preflight` commits under it instead of acquiring (which the live handoff lease blocks outright), and releases it afterward so no phantom lease outlives the terminal `Failed` row. New RED→GREEN test `resume_execution_leased_rejects_undeclared_port_and_still_marks_failed`: with the fix reverted the row stays `Created` (version 0); with the fix it is durably `Failed` and the lease is released. The Acquire path behavior is unchanged (452 engine tests green).

**🟠 Major — foreign-tenant rejection must prove no writes** (all three conformance suites) — fixed in 217a1343: after asserting `ClaimSuperseded`, each suite retries the original handoff with its original scope and claim and asserts `Accepted` — a backend that leased or acknowledged before refusing would fail the retry.

**🟠 Major — claim-token/execution binding coverage for the SQL backends** — fixed in 217a1343: `a_claim_token_cannot_be_paired_with_another_execution` now runs on SQLite and live PostgreSQL too, asserting both sides stay untouched (claimed row acknowledgeable, unrelated execution unleased). All 15 handoff conformance cases pass on in-memory + SQLite + live PostgreSQL.

**🟠 Major (round 1) — fence generation `unwrap_or_default()` in both SQL adapters** — fixed in 5a89e76a: an out-of-u64-range `fencing_generation` now fails closed with `StorageError::Internal` instead of minting fence 0.

**🟡 Minor (round 1) — copied reclaim-predicate comment in the PostgreSQL suite** — fixed in 217a1343 (now describes Postgres's `processed_at_ms < cutoff` predicate).

Round-1 items already closed by earlier commits on this branch: claim-row↔execution binding in the SQL adapters (2e44991f), `ClaimSuperseded` for a gone row in the in-memory adapter (matching SQL semantics), positive lease duration via the `[1s, 24h]` clamp in all three backends, execution-scoped backdate + unique-plugin isolation in the PostgreSQL suite.

